PR #2, opened 2015-04-24
[New Feature]Add parameter to disable highlight auto-detect.

hcoona:
To add a new feature disable highlight auto-detect.

See the issue from [hexo  #1224](https://github.com/hexojs/hexo/issues/1224)

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -9,13 +9,14 @@ module.exports = function(str, options){
 
   if (!hljs) initHighlight();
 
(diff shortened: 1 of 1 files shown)


lzjqsdd:
I had seen what you changed,just when the lang is empty,change its value to 'plain'.Good idea!

tommy351:
I think the auto-detect can be disabled by default. It can make the performance better.

hcoona:
If you make it disable by default, user may notice that their existing codes are not highlighted after a upgrading. You must generate a warning to make sure all users notice this change.

PR #3, opened 2015-09-16
WIP: Is this correct way how to chain promises?

mcepl:
The whole point of this patch is sl.setLimit(20); which should limit the
number of the concurrently running processes to 20.

Fixes https://github.com/hexojs/hexo/issues/1494


leesei:
This *MAY* improve performance but not tested.
Closing for now.

PR #16, opened 2016-08-16
change divs in code blocks back to spans

greg-js:
For some reason I changed the spans to divs in a PR six months ago. Here's the fix to get rid of those divs again. Thanks to @csarron for letting me know the issue

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -32,10 +32,10 @@ function highlightUtil(str, options) {
   for (var i = 0, len = lines.length; i < len; i++) {
     line = lines[i];
     if (tab) line = replaceTabs(line, tab);
-    numbers += '<div class="line">' + (firstLine + i) + '</div>';
-    content += '<div class="line';
+    numbers += '<span class="line">' + (firstLine + i) + '</span><br>';
+    content += '<span class="line';
     content += (mark.indexOf(firstLine + i) !== -1) ? ' marked' : '';
-    content += '">' + line + '</div>';
+    content += '">' + line + '</span><br>';
   }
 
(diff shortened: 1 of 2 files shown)


ahaasler:
I think this is what causes the empty lines error explained in meteor/hexo-theme-meteor#30 and discussed in hexojs/hexo#858 and hexojs/hexo#795.

It's easily fixed the way @tommy351 explains in hexojs/hexo-theme-landscape@ad0c9e95a13d33c12ffa257d44047b38a38fed6d, but I think this isn't something that themes should handle, being this a change all themes should make.

Quoting @0ssifrage:

> Adding css style for `.line` may be a solution, but it's hard to let all the themes change their css files.

ahaasler:
I've been trying the change proposed by @tommy351 and the empty lines are not being copied with a Ctrl-C in the browser, because there's no content.

```
Line 1
Line 3
```

With this PR the empty lines are properly being copied by the browser:

```
Line 1

Line 3
```

I've also tested CSS content without success, because the CSS generated content is not selectable in a browser.

```css
.line:empty:after {
  content: '\000a';
}
```

It would be great if this PR could be merged and released.

stubailo:
Any chance this could be merged? We have to pin to Hexo 3.2.0 in all of our sites because of this issue.

poiru:
@NoahDragon Could we merge this?

NoahDragon:
@poiru Thank for poking me. I will merge this issue.

poiru:
@NoahDragon Thank you! Any Would you also be able to release a new version of `hexo-util`?

PR #22, opened 2017-03-02
BugFix: Correct highlighting of multi-line element (fix #10 and hexojs/hexo#2291)

### LINKED ISSUE #10: Highlighting breaks multiline statements
Say we have a code block in a language that supports multiline comments or strings, such as Coffeescript:

``` coffeescript
myFunc """
  This is a
  string that
  spans multiple
  lines.
"""
```

Hexo will render the following html:

``` html
<figure class="highlight coffeescript">
  <table>
    <tbody>
      <tr>
        <td class="code">
          <pre>
            <span class="line">myFunc <span class="string">"""</span><br>
            <span class="line">  This is a</span><br>
            <span class="line">  string that</span><br>
            <span class="line">  spans multiple</span><br>
            <span class="line">  lines.</span><br>
            <span class="line">"""</span></span><br>
          </pre>
        </td>
      </tr>
    </tbody>
  </table>
</figure>
```

This seems to prevent hljs from properly inserting `<span>` elements. I'll admit I don't understand what [this line](https://github.com/hexojs/hexo-util/blob/master/lib/highlight.js#L36), which adds the `.line` spans, is supposed to accomplish (ie, why do we want each line wrapped in its own span in the first place), but it's doing it in a naive way that doesn't account for existing `span` elements that span more than one line. What it should produce is:

``` html
<figure class="highlight coffeescript">
  <table>
    <tbody>
      <tr>
        <td class="code">
          <pre>
            <span class="line">myFunc <span class="string">"""</span></span><br>
            <span class="line string">  This is a</span><br>
            <span class="line string">  string that</span><br>
            <span class="line string">  spans multiple</span><br>
            <span class="line string">  lines.</span><br>
            <span class="line string">"""</span><br>
          </pre>
        </td>
      </tr>
    </tbody>
  </table>
</figure>
```

...which would be correctly highlighted.

seaoak:
Correct highlighting of multi-line element (fix #10 and hexojs/hexo#2291).

This patch is similar to the patch hexojs/hexo#904 that fixed the issue hexojs/hexo#902.
But the API `highlightByLine()` is no longer supported by [highlight.js](http://highlightjs.readthedocs.io/en/latest/api.html).
So I use the API `highlight()` with continuation.

And also add HTML validation to each test case.
If this validation was implemented, the issue #10 might be detected.
For this puropose, add a new `devDependencies` for `html-tag-validator` package.    
https://www.npmjs.com/package/html-tag-validator    
https://github.com/codeschool/htmlTagValidator

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -94,12 +94,16 @@ function highlight(str, options) {
   var lang = options.lang;
   var autoDetect = options.hasOwnProperty('autoDetect') ? options.autoDetect : false;
 
-  if (!lang) {
-    if (autoDetect) {
-      loadAllLanguages();
-      return hljs.highlightAuto(str);
-    }
+  if (!lang && autoDetect) {
+    loadAllLanguages();
+    lang = (function() {
+      var result = hljs.highlightAuto(str);
+      if (result.relevance > 0 && result.language) return result.language;
+      return;
+    })();
+  }
 
+  if (!lang) {
     lang = 'plain';
   }
 
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -122,7 +122,18 @@ function highlight(str, options) {
 
 function tryHighlight(str, lang) {
   try {
-    return hljs.highlight(lang, str);
+    var matching = str.match(/(\r?\n)/);
+    var separator = matching ? matching[1] : '';
+    var lines = matching ? str.split(separator) : [str];
+    var result = hljs.highlight(lang, lines.shift());
+    var html = result.value;
+    while (lines.length > 0) {
+      result = hljs.highlight(lang, lines.shift(), false, result.top);
+      html += separator + result.value;
+    }
+
+    result.value = html;
+    return result;
   } catch (err) {
     return;
   }
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "chai": "^3.5.0",
     "eslint": "^2.12.0",
     "eslint-config-hexo": "^1.0.3",
+    "html-tag-validator": "^1.5.0",
     "istanbul": "^0.4.3",
     "jscs": "^3.0.4",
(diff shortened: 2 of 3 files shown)


stevenjoezhang:
The `continuation` option is deprecated since highlight.js verion 10.7.1

See also https://github.com/hexojs/hexo-util/issues/245 https://github.com/highlightjs/highlight.js/issues/2277

joshgoebel:
The "blessed" way to do line by line highlighting (answer is mine):

https://stackoverflow.com/questions/64280814/how-can-i-correctly-highlight-a-line-by-line-code-using-highlight-js-react

IE, you highlight the whole content in full and then go back over it looking for the line-breaks and rewriting the HTML to open/close tags at the beginning/end of each line.  The HTML we generate is very clean (all `<>` are escaped), so one could do this with a regex to match HTML tags but I'd take a look at the `__emitter` API (as suggested in my answer) if you don't require 100% API stability guarantees (`__emitter`is technically a very stable private API, but technically it could break in-between major releases).

If you require 100% API stability then you'd need to parse the resulting HTML and fix it up manually.

Also, FYI: There are many ways to avoid dealing with individual lines entirely with creative use of CSS, though that can't cover all scenarios and I'm not sure about this use case.

joshgoebel:
There is probably a very minimal approach to parsing the raw HTML that would go something like:

```js
lines = hljs.highlight(code).value.split("\n")
open = []
lines = lines.map((line) => {
  out = openTags(open) + line;
  tags = line.matchAll(HTML_TAG_RE)
  tags = removeAlreadyClosedTags(tags)
  open = [...tags]
  return out + closeTags(tags.length)
});
```

stevenjoezhang:
@joshgoebel Thanks for your suggestion, I implemented this function (parse the resulting HTML and fix it up) using a stack: https://github.com/hexojs/hexo-util/pull/246

PR #29, opened 2017-11-02
Apply hljs classPrefix to container figure element

Krakob:
Building on #15. As hljs uses 'hljs' by default for its containers, and not 'highlight' as hexo currently has hardcoded, this makes sure that the original hljs stylesheets maintain compatibility, should you use hljs as your parameter.

It should be noted that a `-` is appended to the class prefix so that we get, for example, `hljs` for the container and `hljs-number`. Also added some quick documentation as was requested in #15.

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -5,13 +5,19 @@ var Entities = require('html-entities').XmlEntities;
 var entities = new Entities();
 var alias = require('../highlight_alias.json');
 
-hljs.configure({
-  classPrefix: ''
-});
+// hljs.configure({
+//   classPrefix: ''
+// });
 
 function highlightUtil(str, options) {
   if (typeof str !== 'string') throw new TypeError('str must be a string!');
   options = options || {};
+
+  var classPrefix = options.hasOwnProperty('classPrefix') ? options.classPrefix : '';
+
+  hljs.configure({
+    classPrefix: classPrefix
+  });
 
   var gutter = options.hasOwnProperty('gutter') ? options.gutter : true;
   var wrap = options.hasOwnProperty('wrap') ? options.wrap : true;
@@ -20,12 +26,6 @@ function highlightUtil(str, options) {
   var mark = options.hasOwnProperty('mark') ? options.mark : [];
   var tab = options.tab;
   var data = highlight(str, options);
-  
-  if (options.hasOwnProperty('classPrefix')) {
-    hljs.configure({
-      classPrefix: options.classPrefix
-    });
-  }
 
   if (!wrap) return data.value;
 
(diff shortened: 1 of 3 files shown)


Krakob:
Not quite sure what's up with the test, but it seems to try to close some spans twice, e.g.

`a = <span class="hljs-string">`</span><br>` expected,
`a = <span class="hljs-string">`</span></span><br>` received

Can't say I know unit testing very well.

NoahDragon:
I will have a look.

lib/highlight.js
```diff
@@ -5,14 +5,21 @@ var Entities = require('html-entities').XmlEntities;
 var entities = new Entities();
 var alias = require('../highlight_alias.json');
 
-hljs.configure({
-  classPrefix: ''
-});
+// hljs.configure({
+//   classPrefix: ''
+// });
```
NoahDragon: Could you please remove the comment out codes?

Krakob: Done :)

lib/highlight.js
```diff
@@ -5,14 +5,17 @@ var Entities = require('html-entities').XmlEntities;
...
-hljs.configure({
-  classPrefix: ''
-});
-
 function highlightUtil(str, options) {
   if (typeof str !== 'string') throw new TypeError('str must be a string!');
   options = options || {};
 
+  var classPrefix = options.hasOwnProperty('classPrefix') ? options.classPrefix + '-' : '';
+  var containerPrefix = options.hasOwnProperty('classPrefix') ? options.classPrefix : 'highlight';
```
NoahDragon: Shouldn't we remove this line?

JLHwung:
The unit-test failed on `classPrefix`, could you give it a fix?

Krakob:
I can't figure out how :(

The feature was working in the initial PR, as seen here: https://travis-ci.org/hexojs/hexo-util/jobs/146592708

Yet this commit failed: https://github.com/hexojs/hexo-util/pull/29/commits/909523ad3e198c24cc7ff9a572ce7fdec02e775d

And the failure reason being, as mentioned, generating duplicate closing tags for some reason, which sounds like an internal issue in hljs. The only line in the lib file we can link directly to hljs is 16, but it's compensated for in the test as the hyphen in classPrefix was removed.

NoahDragon:
@Krakob The config for classPrefix in highlight.js seems changed the behavior of generating lines.

Expected a line:
```html
<span class="line">
    <span class="hljs-string">  Multi</span>
</span>
<br>
```

The line generated via higlight.js with the classPrefix:
```html
<span class="line">  Multi</span>
<br>
```

Not sure how this happens, will investigate deeper.

JLHwung:
closed in favor of #30.

PR #33, opened 2018-04-08
Add David Badge from README.md

segayuu:


segayuu:
thanks @JLHwung !
I fixed it from CRLF to LF!

PR #35, opened 2018-07-09
Refactoring es2016nify

segayuu:


segayuu:
travis ci is causing an error is an error specific to the test-cov command.
My development environment is windows, the test command did not detect anything other than windows specific errors.

segayuu:
maybe issue https://github.com/gotwarlost/istanbul/issues/717 ...

lib/highlight.js
```diff
@@ -1,72 +1,73 @@
...
+  lines.forEach((line, i) => {
     if (tab) line = replaceTabs(line, tab);
-    numbers += '<span class="line">' + (firstLine + i) + '</span><br>';
+    numbers += `<span class="line">${firstLine + i}</span><br>`;
     content += formatLine(line, firstLine + i, mark, options);
-  }
+  });
 
-  result += '<figure class="highlight' + (data.language ? ' ' + data.language : '') + '">';
+  let result = `<figure class="highlight${language ? ` ${language}` : ''}">`;
```
tcrowe: In the future maybe we can go into some of these ternary expressions and make them more explicit. That was there before so I didn't want to block it for this PR.

segayuu:
@JLHwung
https://coveralls.io/builds/16686048/source?filename=lib%2Fhighlight.js
As a result of comparison with the status page of the corresponding file in the master branch, the reason why the coverage rate fell down is probably because the number of lines has been reduced.

lib/highlight.js
segayuu: @tcrowe thank review!
As you can see, I set it explicitly.

tcrowe:
This will bring `./lib/hash.js` coverage to 100%

```js
...
it('HashStream', () => {
  const content1 = '123456';
  const content2 = '654321';
  const stream = new hash.HashStream();

  // explicit convert
  stream.write(Buffer.from(content1));

  // implicit convert
  stream.write(content2);
  stream.end();

  stream.read().should.eql(sha1(content1 + content2));
});
...
```

lib/cache_stream.js
```diff
@@ -11,7 +11,7 @@ function CacheStream() {
 require('util').inherits(CacheStream, Transform);
 
 CacheStream.prototype._transform = function(chunk, enc, callback) {
-  var buf = chunk instanceof Buffer ? chunk : Buffer.from(chunk, enc);
+  const buf = chunk instanceof Buffer ? chunk : Buffer.from(chunk, enc);
```
tcrowe: I can't get this `Buffer.from` branch to run in the test. Can it ever be something other than Buffer? If not the `Buffer.from` branch can be removed and coverage will be 100%.

tcrowe: Or you can do something tricky like this in `./test/cache_stream.js`

```js
it('default', () => {
  const src = new Readable();
  const cacheStream = new CacheStream();
  const content1 = Buffer.from('test1');
  const content2 = 'test2';
  let transformed = false;

  src.push(content1);
  src.push(null);
  src.pipe(cacheStream);

  cacheStream._transform(content2, 'utf8', () => { transformed = true; });

  cacheStream.on('finish', () => {
    cacheStream.getCache().should.eql(content1 + content2);
    transformed.should.be.true;
  });
});
```

segayuu: https://nodejs.org/api/stream.html#stream_transform_transform_chunk_encoding_callback
Even if stream is not an objectMode, it is possible to pass string. For this reason, we need to assume that string will flow.

segayuu: As a result of continuing committing, we found that `chunk` will be string only when `cacheStream.setEncoding('utf8')` is done. In `src.setEncoding('utf 8')`, `chunk` in `_transform()` was converted to `Buffer`.

segayuu: The type of `chank` of `_transform()`
- Set `decodeStrings` of the` Transform (super: Writeble) `class's constructor option to` false`.
- Calling `setEncoding()` on the source `Readable`.
Only when it meets the above conditions, it is `string`, otherwise it is `Buffer`.

segayuu: After finishing the previous behavior check, in the case of the current implementation, we were able to grasp that `chunk` is only `Buffer` ...

tcrowe:
@segayuu It's looking good. Whenever you feel like it's finished let me know and I'll test, review, & merge.

segayuu:
@tcrowe thank!
https://nodejs.org/api/crypto.html#crypto_class_hash
https://github.com/nodejs/node/blob/v10.6.0/lib/internal/crypto/hash.js#L42
Hash class is not-objectMode Transform stream class.
The only difference from HashStream is that objectMode is not set.

https://github.com/segayuu/hexo-util/blob/refactoring-from-lebab/lib/hash.js#L19
However, the current implementation of hashStream#_transform() does not assume that object will come.

For this reason, I am worried about whether to implement it so that it can deal with object or worry about deleting implementation other than algorithm as a useless wrapper class.

tcrowe:
`util.HashStream` is only used in two places https://github.com/hexojs/hexo/search?q=hashstream&unscoped_q=hashstream

Only `String` or `Buffer` go into it.

What do you think we should do?

segayuu:
For me, the idea is to make the current HashStream a deprecated class and create a new createSha1Hash ().

tcrowe:
@segayuu Okay cool! 🤓

segayuu:
@tcrowe It would be greatly appreciated if you could do a second review.

tcrowe:
@segayuu Okay! It might be when I wake up in the morning. It's late here now.

PR #36, opened 2018-07-31
Allows empty attributes in html tag helper

Xaviju:
Fixes this issue -> https://github.com/hexojs/hexo/issues/3222

### DIFF
--- a/lib/html_tag.js
+++ b/lib/html_tag.js
@@ -6,7 +6,10 @@ function htmlTag(tag, attrs, text) {
   var result = '<' + tag;
 
   for (var i in attrs) {
-    if (attrs[i]) result += ' ' + i + '="' + attrs[i] + '"';
+    if (
+      attrs[i] !== undefined
+      && attrs[i] !== null
+    ) result += ' ' + i + '="' + attrs[i] + '"';
   }
 
   result += text == null ? '>' : '>' + text + '</' + tag + '>';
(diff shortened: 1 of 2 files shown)


lib/html_tag.js
```diff
@@ -6,7 +6,7 @@ function htmlTag(tag, attrs, text) {
   var result = '<' + tag;
 
   for (var i in attrs) {
-    if (attrs[i]) result += ' ' + i + '="' + attrs[i] + '"';
+    if (attrs[i] || attrs[i] === '') result += ' ' + i + '="' + attrs[i] + '"';
```
JLHwung: I suggest we use stricter `attrs[i] !== undefined && attrs[i] !== null` on this check.

Xaviju: There is an error already if an attribute is not defined.  Should I add the test?
```
it('tag + empty attr', function() {
    try {
      htmlTag('img', {
        src: 'http://placekitten.com/200/300',
        hidden
      });
    } catch (err) {
      err.should.have.property('message', 'hidden is not defined');
    }
  });
```
or
```
it('tag + empty attr', function() {
    try {
      htmlTag('img', {
        src: 'http://placekitten.com/200/300',
        alt: undefined
      });
    } catch (err) {
      err.should.have.property('message', 'alt is not defined');
    }
  });
```

Xaviju:
Thinking it twice, just adding the conditions you suggested, all possible valid attributes are working as expected. I've added a few tests to ensure that works properly.

NoahDragon:
Could you please fix the error in travis-ci?

Xaviju:
@NoahDragon should be fixed now

NoahDragon:
@Xaviju Great, thank you.

PR #37, opened 2018-08-11
Fix sublanguages highlight

l-vo:
Highlightjs has a sublanguages feature which allows some languages to incorporate and highlight another language inside it. When I use a sublanguage in Hexo, it doesn't work:

without_highlight

This PR make the sublanguages functional:

highlight

The sublanguage feature requires that the sublangage definition is loaded. But Hexo only load the definition of the main language of the code chunk. I suggest to always load all language definitions, like it's done in highlightjs library and in hexo-util tests (so the tests are currently not totally right since the definitions loaded in tests are not the same as the definitions actually loaded by hexo-utils).

Also, it could be possible to dynamically load sublanguages but it will make hexo-util too closely coupled with hightlighjs IMHO.

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var hljs = require('highlight.js/lib/highlight');
+var hljs = require('highlight.js');
 var Entities = require('html-entities').XmlEntities;
 var entities = new Entities();
 var alias = require('../highlight_alias.json');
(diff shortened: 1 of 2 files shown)


segayuu:
This PR is loosely coupled or a trade-off with memory, and it is difficult to decide which one to prioritize.
Whether this behavior should be changed as hexo-util libraries remains in doubt.
However, I acknowledge that it is not good to use sub language features at Hexo in the current situation.

PR #44, opened 2018-10-18
Refactoring es2016nify

segayuu:
This PR is divided PR because #35 has become big.

### DIFF
--- a/lib/cache_stream.js
+++ b/lib/cache_stream.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var Transform = require('stream').Transform;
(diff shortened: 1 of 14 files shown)


YoshinoriN:
👋 @segayuu

Btw, I need a lot of effort for review that like this pull-request .

At first, a lot of file and code changed.
Second, some different meanings changes include this pull-request.

forexamle

* var to let/const
* use defineProperty instead of `__define__`
* use allow function

...etc

So, I could have approved more faster, if these changes were divided into some pull requests. 😆

PR #58, opened 2019-06-10
Add Color class from tagcloud hexo helper

segayuu:
## Agenda

`tagcloud` helper depends on the `Color` inner class. This class is reusable but is not currently tested alone.
Therefore, after porting to `hexo-util`, we separate tagcloud-dependent code from the original test. This improves the maintainability.

## Clone

```
$ git clone https://github.com/segayuu/hexo-util
$ cd hexo-util
$ git fetch
$ git checkout Add-Color
```

## Test

```
$ npm install
$ npm run test
```

## Related
https://github.com/hexojs/hexo/blob/3.8.0/lib/plugins/helper/tagcloud.js
https://github.com/hexojs/hexo/blob/3.8.0/test/scripts/helpers/tagcloud.js

### DIFF
--- a/lib/color.js
+++ b/lib/color.js
@@ -216,7 +216,7 @@ class Color {
   _parse(color) {
     color = color.toLowerCase();
 
-    if (colorNames.hasOwnProperty(color)) {
+    if (Object.prototype.hasOwnProperty.call(colorNames, color)) {
       const obj = colorNames[color];
 
       this.r = obj.r;
--- a/lib/color.js
+++ b/lib/color.js
@@ -1,0 +1,336 @@
+'use strict';
+
+// https://github.com/imathis/hsl-picker/blob/master/assets/javascripts/modules/color.coffee
+const rHex3 = /^#[0-9a-f]{3}$/;
+const rHex6 = /^#[0-9a-f]{6}$/;
+const rRGB = /^rgba?\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,?\s*(0?\.?\d+)?\s*\)$/;
+const rHSL = /^hsla?\(\s*(\d{1,3})\s*,\s*(\d{1,3})%\s*,\s*(\d{1,3})%\s*,?\s*(0?\.?\d+)?\s*\)$/;
+
+// https://www.w3.org/TR/css3-color/#svg-color
+const colorNames = {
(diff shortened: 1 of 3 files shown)


tcrowe:
That looks interesting, @segayuu! Is it possible you can add more information? Maybe also provide the code to clone & test?

segayuu:
Thank @tcrowe !

> That looks interesting, @segayuu! Is it possible you can add more information? Maybe also provide the code to clone & test?

The PR text has been added (though it was originally 0 characters).

YoshinoriN:
@tcrowe 
It's just migrate from current [hexo](https://github.com/hexojs/hexo/blob/3.8.0/lib/plugins/helper/tagcloud.js) source code.

@segayuu 
My understanding, we need following procedure. Is it correct?

1. Merge this PR
2. Publish new hexo-util
3. Update hexo's hexo-util dependency & delete current [hexo](https://github.com/hexojs/hexo/blob/3.8.0/lib/plugins/helper/tagcloud.js) source code.
4. Announce breaking change & migration guide for users.

YoshinoriN:
Btw, reported an issue about `tagcloud` https://github.com/hexojs/hexo/issues/3622
We might be better to investigate it before merge this.

segayuu:
Thank @YoshinoriN !
The procedure you suggested is basically correct, but only the fourth one is unnecessary.
The reason is that the `Color` class before PR is an inner class, is not exposed as a module, and has no way to access it (this is one of the purpose of PR).
From the user's point of view, it is not a Breaking Change, as it just added an API(it is necessary to fix it as a bug if its behavior changes).

> Btw, reported an issue about tagcloud hexojs/hexo#3622
We might be better to investigate it before merge this.

I did not know the issue, thank you!
However, this is the code related to the `tagcloud()` and the `Hexo` class, not to the `Color` class itself.

YoshinoriN:
@segayuu 
IMHO we might as well include this PR to hexo v4. How do you think?
If yes, we should review & merge this PR & release a new version of hexo-util before release hexo v4.

> https://github.com/hexojs/hexo/issues/3508

segayuu:
@YoshinoriN I agree with your opinion! 😆

tomap:
@hexojs/core anyone wants to add something or we can merge?

curbengh:
I suggest to rebase to include test in node 8-12.

segayuu:
Since rebase was done and a new error was corrected, there was no problem with merge.

PR #66, opened 2019-07-28
Compatible with XHTML syntax

stevenjoezhang:


tomap:
Why would you need to be compatible with xhtml, now that everyone is using html5?

stevenjoezhang:
To determine how many Hexo themes are using html5, I randomly git cloned about 20 themes from https://github.com/hexojs/site/blob/master/source/_data/themes.yml, search result shows that they all contain `<!doctype html>` in `layout`
If you need accurate numbers, you can write a script / spider to parse the file.

In addition, html4 is also compatible with xhtml syntax. See also https://caniuse.com/#search=xhtml

tomap:
So you are telling me that you found 100% html5. So why do you need xhtml syntax?

stevenjoezhang:
> now that everyone is using html5?

Emm, I thought you meant that very few people are using html5..

stevenjoezhang:
> So why do you need xhtml syntax?

Because I found that Hexo's meta_generator uses the XHTML syntax. This leads to the presence of mixed HTML5 and XHTML in Hexo. After Hexo 4 fixed the meta_generator, this PR is no longer necessary.

https://github.com/hexojs/hexo/commit/651f34bf6d2be5cb9c07009e857a73ecad3b414c

PR #67, opened 2019-08-04
fix: prevent firstLine being parsed as a string

curbengh:
firstLine variable sometimes can be parsed as a string.
let say, `firstLine = 1`; `firstLine + 3` can results in `13`, instead of expected `4`.

See https://github.com/hexojs/hexo/pull/3646


curbengh:
It's not supposed to be parsed as a string though,
```js
const firstLine = 1;
const i = 2;
const check = `<span class="line">${firstLine + i}</span><br>`
console.log(check)
```
works as expected.

curbengh:
My guess is that hexo parse output of hexo-util as string, maybe too aggressively.

PR #57, opened 2019-02-25
Bump version to 1.0.0 following 1 year of changes

tomap:
to show that it is stable

PR #69, opened 2019-08-09
fix: fs.write instead of stdout

curbengh:
in some edge case, stdout can [be empty](https://stackoverflow.com/questions/54190596/no-console-log-to-stdout-when-running-npm-test-jest) when running npm script, i.e. `npm run prepare`. `fs.write` is more reliable.

"highlight_alias.json" should be published, it's generated when `npm publish`. The value in "files" attribute is needed even when `"hexo-util": "hexojs/hexo-util"`, otherwise it would not be generated when `npm install`.

Fixes https://github.com/hexojs/hexo/pull/3646#issuecomment-519465750, Address https://github.com/hexojs/hexo-util/pull/57#issuecomment-519031638

### How to test
`$ rm -r node_modules/`

``` diff
package.json
-  "hexo": "^3.9.0"
+  "hexo": "curbengh/hexo#rc"
```

``` bash
$ npm i
$ hexo -v # should be no error
$ npm test
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "eslint": "eslint .",
     "test": "mocha test/index.js",
     "test-cov": "nyc npm run test",
-    "build:highlight": "node scripts/build_highlight_alias.js > highlight_alias.json",
+    "build:highlight": "node scripts/build_highlight_alias.js",
     "prepare": "npm run build:highlight"
   },
   "directories": {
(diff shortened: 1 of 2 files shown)


curbengh:
I'll es6-fy scripts/build_highlight_alias.js in a separate PR.

PR #75, opened 2019-08-29
fix: export Color function

curbengh:
Currently, using the Color function requires `const Color = require('../../../node_modules/hexo-util/lib/color');`.

It should be just `const Color = require('hexo-util').Color;`.


curbengh:
I'll publish this as v1.0.1.

PR #77, opened 2019-09-02
feat: add encodeURL()

curbengh:
This PR is stemmed from IDN parsing issue (https://github.com/hexojs/hexo-generator-feed/issues/63). It was later (partially) fixed in https://github.com/hexojs/hexo-generator-feed/pull/82, without encodeURI the root value. Full fix is later proposed in https://github.com/hexojs/hexo-generator-feed/pull/88.

IDN incompatibility is also found in [hexo-generator-sitemap](https://github.com/hexojs/hexo-generator-sitemap/pull/66) and [open_graph()](https://github.com/hexojs/hexo/pull/3686).

All the proposed fixes actually share similar function, so I propose to put it in hexo-util instead.

### DIFF
--- a/lib/encode_url.js
+++ b/lib/encode_url.js
@@ -1,0 +1,17 @@
+'use strict';
+
+const { parse, format } = require('url');
+
+function encodeURL(str) {
+  if (parse(str).protocol) {
+    return format({
+      protocol: parse(str).protocol,
+      hostname: parse(str).hostname,
+      pathname: encodeURI(decodeURI(parse(str).pathname))
+    });
+  }
+
+  return encodeURI(str);
+}
+
+module.exports = encodeURL;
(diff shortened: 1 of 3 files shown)


lib/encode_url.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const { parse, format } = require('url');
+
+function encodeURL(str) {
+  if (parse(str).protocol) {
+    return format({
+      protocol: parse(str).protocol,
+      hostname: parse(str).hostname,
+      pathname: encodeURI(decodeURI(parse(str).pathname))
```
curbengh: decode+encode is necessary to avoid double-encode the space character; the space character is  encoded by url api by default.

curbengh:
@tomap what do you think, in relation to the IDN issue?

tomap:
Looks good. unit tests seem to cover all cases

Issue #83, opened 2019-09-11
encodeURL() will lose hash in URL


SukkaW:
```javascript
const { encodeURL } = require("hexo-util");
console.log(encodeURL('https://example.com/blog/#test'));
// "https://example.test/blog/"
```

```javascript
const { encodeURL } = require("hexo-util");
encodeURL('https://example.com/#test');
// "https://example.com/"
```

https://github.com/hexojs/hexo-util/blob/a3abc361caec3f1c09108141c23f2d813d0d387d/lib/encode_url.js#L10

`pathname` doesn't include `hash`. `parse(str).href` should be used in this case.

@curbengh

PR #84, opened 2019-09-11
feat(full_url_for): bring up a new helper

SukkaW:
https://github.com/hexojs/hexo/pull/3701

### DIFF
--- a/lib/full_url_for.js
+++ b/lib/full_url_for.js
@@ -11,7 +11,6 @@ function fullUrlForHelper(path = '/') {
   // Exit if this is an external path
   if (data.protocol) return path;
 
-  // Prepend root path
   path = encodeURL(config.url + `/${path}`.replace(/\/{2,}/g, '/'));
   return path;
 }
--- a/lib/full_url_for.js
+++ b/lib/full_url_for.js
@@ -1,0 +1,19 @@
+'use strict';
+
+const { parse } = require('url');
+const encodeURL = require('./encode_url');
+
+function fullUrlForHelper(path = '/') {
+  if (path.startsWith('//')) return path;
+  const { config } = this;
+  const data = parse(path);
+
+  // Exit if this is an external path
+  if (data.protocol) return path;
+
+  // Prepend root path
+  path = encodeURL(config.url + `/${path}`.replace(/\/{2,}/g, '/'));
+  return path;
+}
+
+module.exports = fullUrlForHelper;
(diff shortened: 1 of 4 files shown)


SukkaW:
The test will passed after #83 is fixed.

curbengh:
The test pass now, just need to remove `console.log` and line ending.

lib/full_url_for.js
```diff
@@ -0,0 +1,19 @@
...
+
+function fullUrlForHelper(path = '/') {
+  if (path.startsWith('//')) return path;
+  const { config } = this;
+  const data = parse(path);
+
+  // Exit if this is an external path
+  if (data.protocol) return path;
+
+  // Prepend root path
```
curbengh: I don't think this comment is necessary, the next line is obvious on what it does.

test/full_url_for.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+describe('full_url_for', () => {
+  const ctx = {
+    config: {}
+  };
+
+  const fullUrlFor = require('../lib/full_url_for').bind(ctx);
+
+  it('internal url - root directory', () => {
+    ctx.config.root = '/';
```
curbengh: `config.root` is never used, so there's no need to test it.

test/full_url_for.spec.js
```diff
@@ -0,0 +1,44 @@
...
+  it('internal url - subdirectory', () => {
+    ctx.config.root = '/blog/';
+    ctx.config.url = 'https://example.com/blog';
+    fullUrlFor('index.html').should.eql(ctx.config.url + '/index.html');
+    fullUrlFor('/').should.eql(ctx.config.url + '/');
+  });
+
+  it('internal url - no duplicate slash', () => {
+    ctx.config.root = '/';
+    ctx.config.url = 'https://example.com';
```
curbengh: `config.root` is never used, it should be `ctx.config.url = 'https://example.com/';`

lib/full_url_for.js
SukkaW: Removed.

test/full_url_for.spec.js
SukkaW: Removed.

test/full_url_for.spec.js
SukkaW: Removed.

PR #85, opened 2019-09-12
fix: handle url with hash, query, auth and port

### LINKED ISSUE #83: encodeURL() will lose hash in URL

curbengh:
Fixes #83

### DIFF
--- a/lib/encode_url.js
+++ b/lib/encode_url.js
@@ -5,8 +5,9 @@ const { parse, format } = require('url');
 function encodeURL(str) {
   if (parse(str).protocol) {
     const obj = Object.assign({}, {
+      auth: parse(str).auth,
       protocol: parse(str).protocol,
-      hostname: parse(str).hostname,
+      host: parse(str).host,
       pathname: encodeURI(decodeURI(parse(str).pathname))
     });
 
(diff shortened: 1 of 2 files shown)


curbengh:
The [URL API](https://nodejs.org/api/url.html#url_legacy_url_api) mentions `path` includes pathname and query, but
```format({ path: encodeURI(decodeURI(parse(str).path)) })```
doesn't output path nor query, thus `pathname` is used instead.

File: lib/encode_url.js [hunk 1]
```diff
@@ -4,11 +4,21 @@ const { parse, format } = require('url');
 
 function encodeURL(str) {
   if (parse(str).protocol) {
-    return format({
+    const obj = Object.assign({}, {
       protocol: parse(str).protocol,
       hostname: parse(str).hostname,
```
SukkaW: Should it use `host` instead of `hostname`? `host` includes port while `hostname` doesn't.

And `auth` (`user:pass`) might be needed in output result (without encoded) as well.

curbengh: Fixed

PR #88, opened 2019-09-15
docs: url_for option

### LINKED ISSUE #87: Docs: url_for options
https://github.com/hexojs/hexo-util/blob/e1558933f28313c5e85c6473e81d1da8a7d5b9a5/lib/url_for.js#L7

hexo-util `url_for()` has an `options` and the usage should be included in README.

The hexo documents need update as well.

curbengh:
also update test to be more obvious on the override option

Closes #87


curbengh:
I'll update hexojs/site after this is merged.

PR #92, opened 2019-09-15
fix(encodeURL): encode path once only

curbengh:


lib/encode_url.js
```diff
@@ -22,7 +22,7 @@ function encodeURL(str) {
     return format(obj);
   }
 
-  return encodeURI(str);
+  return encodeURI(decodeURI(str));
```
dailyrandomphoto: In some edge case, decode a text which is not encoded by `encodeURI`, `decodeURI` will raise `URI malformed` exception.
e.g.
```js
encodeURL("http://foo.com/bar?query=%");
encodeURL("/bar?query=%");
```
```
URIError: URI malformed
      at decodeURI (<anonymous>)
      at encodeURL (lib/encode_url.js:25:20)
```

Use `try, catch` can prevent this exception.
```js
// L19
- Object.assign(obj, { search: encodeURI(decodeURI(parse(str).search)) });
+ Object.assign(obj, { search: encodeURI(safeDecodeURI(parse(str).search)) });
// L25
- return encodeURI(decodeURI(str));
+ return encodeURI(safeDecodeURI(str));


function safeDecodeURI(str) {
  try {
    return decodeURI(str);
  } catch (err) {
    return str;
  }
}
```

curbengh: https://github.com/hexojs/hexo-util/pull/96

PR #93, opened 2019-09-16
fix(html_tag): escape html and encode url by default

curbengh:
Related to https://github.com/hexojs/hexo/pull/3704 (cc @dailyrandomphoto)
This is to transform
``` html
<a href="/posts/test1/" title="this is a title with <a tag>.">this is a text with </a><a tag="">.</a>
```
to
``` html
<a href="/posts/test1/" title="this is a title with &lt;a tag&gt;.">this is a text with &lt;/a&gt;&lt;a tag=""&gt;.</a>
```

There is an option to disable escape _just_ the text.

``` js
htmlTag('a', {href: 'http://foo.com'}, '<b>bold</b> text', false)
```

``` html
<a href="http://foo.com"><b>bold</b> text</a>
```

### DIFF
--- a/lib/html_tag.js
+++ b/lib/html_tag.js
@@ -1,15 +1,25 @@
 'use strict';
 
-function htmlTag(tag, attrs, text) {
+const encodeURL = require('./encode_url');
+const escapeHTML = require('./escape_html');
+
+function htmlTag(tag, attrs, text, escape = true) {
   if (!tag) throw new TypeError('tag is required!');
 
-  let result = `<${tag}`;
+  let result = `<${escapeHTML(tag)}`;
 
   for (const i in attrs) {
-    if (attrs[i] != null) result += ` ${i}="${attrs[i]}"`;
+    if (attrs[i] === null || typeof attrs[i] === 'undefined') result += '';
+    else {
+      if (i === 'href' || i === 'src') result += ` ${i}="${encodeURL(attrs[i])}"`;
+      else result += ` ${escapeHTML(i)}="${escapeHTML(String(attrs[i]))}"`;
+    }
   }
 
-  result += text == null ? '>' : `>${text}</${tag}>`;
+  if (escape && text) text = escapeHTML(String(text));
+
+  if (text === null || typeof text === 'undefined') result += '>';
+  else result += `>${text}</${escapeHTML(tag)}>`
 
   return result;
 }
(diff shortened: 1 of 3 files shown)


curbengh:
Just updated docs.

lib/html_tag.js
```diff
@@ -1,15 +1,25 @@
...
 
-  let result = `<${tag}`;
+  let result = `<${escapeHTML(tag)}`;
 
   for (const i in attrs) {
-    if (attrs[i] != null) result += ` ${i}="${attrs[i]}"`;
+    if (attrs[i] === null || typeof attrs[i] === 'undefined') result += '';
+    else {
+      if (i === 'href' || i === 'src') result += ` ${i}="${encodeURL(attrs[i])}"`;
+      else result += ` ${escapeHTML(i)}="${escapeHTML(String(attrs[i]))}"`;
```
dailyrandomphoto: It's not safe to escape attributes when which value is url.
e.g.
```js
data-url="http://example.com/"
=>
data-url="http:&#x2F;&#x2F;example.com&#x2F;"
```

I think escape `"` is enough.
```js
else result += ` ${escapeHTML(i)}="${String(attrs[i]).replace(/"/g, "&quot;")}"`;
```

```

File: test/html_tag.spec.js [hunk 2]
```diff
@@ -62,4 +62,23 @@ describe('htmlTag', () => {
...
+      bar: '<b>'
+    }, '<baz>').should.eql('<foo bar="&lt;b&gt;">&lt;baz&gt;</foo>');
+  });
+
+  it('escape html tag (escape off)', () => {
+    htmlTag('foo', {
+      bar: '<b>'
+    }, '<baz>', false).should.eql('<foo bar="&lt;b&gt;"><baz></foo>');
+  });
+
```
dailyrandomphoto: Can you add these test cases.
```js
  it('tag + data-attrs', () => {
    htmlTag('foo', {
      'data-url': 'http://example.com/'
    }, '<baz>').should.eql('<foo data-url="http://example.com/">&lt;baz&gt;</foo>');
  });

  it('tag + bad attrs', () => {
    htmlTag('foo', {
      'bar': 'bar" class="badclass'
    }, '<baz>').should.eql('<foo bar="bar&quot; class=&quot;badclass">&lt;baz&gt;</foo>');
  });

  it('nested tags', () => {
    htmlTag('div', {
    	'class': 'parent'
    }, htmlTag('a', {
    	'href': 'http://example.com/'
    }, 'link'), false).should.eql('<div class="parent"><a href="http://example.com/">link</a></div>');
  });
```

PR #94, opened 2019-09-16
release: 1.2.1

curbengh:
to include https://github.com/hexojs/hexo-util/pull/92 #93 

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hexo-util",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "description": "Utilities for Hexo.",
   "main": "lib/index",
   "scripts": {


SukkaW:
#93 If we disable escape by default, will it still cause breaking changes?

Also, many helper might need refactor after this PR (like escape options in `open_graph()` helper):

https://github.com/hexojs/hexo/blob/bc474dca9331dd8f85579bf7c028b7b10b39c0b1/lib/plugins/helper/open_graph.js#L20

dailyrandomphoto:
It's not safe to escape attributes when which value is url.
e.g.
```js
data-url="http://example.com/"
=>
data-url="http:&#x2F;&#x2F;example.com&#x2F;"
```

I left some comments on #93 .

curbengh:
> #93 If we disable escape by default, will it still cause breaking changes?
no, but I encourage escape by default; meaning if users want to embed tag inside tag, they would have to opt out. Alternatively we could publish v1.3; v2 is too drastic though.

> It's not safe to escape attributes when which value is url.
Fixed in #96 by detecting attribute `if (i.includes('href') || i.includes('src') || i.includes('url'))`.

Alternative approach is to detect url `url.parse(str).protocol` but that doesn't cover `/foo/bar`. `url.parse(str).pathname` is not suitable because it covers _all_ string.

curbengh:
Superseded by https://github.com/hexojs/hexo-util/pull/98

PR #96, opened 2019-09-17
fix(html_tag): encode url value of url-related attributes and skip escape/encode <style>

curbengh:
address https://github.com/hexojs/hexo-util/pull/94#issuecomment-531865496 to support url value in data-* attributes (e.g. data-url, [`data-src`](https://cloudinary.com/documentation/responsive_images#step_2_set_the_img_tag_parameters))
cc @dailyrandomphoto

test case is already covered by
``` js
  it('encode url', () => {
    htmlTag('img', {
      src: 'http://foo.com/bár.jpg'
    }).should.eql('<img src="http://foo.com/b%C3%A1r.jpg">');
  });
```

### DIFF
--- a/lib/html_tag.js
+++ b/lib/html_tag.js
@@ -2,7 +2,6 @@
 
 const encodeURL = require('./encode_url');
 const escapeHTML = require('./escape_html');
-const { parse } = require('url');
 
 function htmlTag(tag, attrs, text, escape = true) {
   if (!tag) throw new TypeError('tag is required!');
--- a/lib/html_tag.js
+++ b/lib/html_tag.js
@@ -22,8 +22,8 @@ function htmlTag(tag, attrs, text, escape = true) {
   for (const i in attrs) {
     if (attrs[i] === null || typeof attrs[i] === 'undefined') result += '';
     else {
-      if (i.endsWith('href') || i.endsWith('src') || i.endsWith('url')) result += ` ${i}="${encodeURL(attrs[i])}"`;
-      else if (i.endsWith('srcset')) result += ` ${i}="${encSrcset(attrs[i])}"`;
+      if (i.endsWith('href') || i.endsWith('src') || i.endsWith('url')) result += ` ${escapeHTML(i)}="${encodeURL(attrs[i])}"`;
+      else if (i.endsWith('srcset')) result += ` ${escapeHTML(i)}="${encSrcset(attrs[i])}"`;
       else result += ` ${escapeHTML(i)}="${escapeHTML(String(attrs[i]))}"`;
     }
   }
--- a/lib/html_tag.js
+++ b/lib/html_tag.js
@@ -2,6 +2,7 @@
 
 const encodeURL = require('./encode_url');
 const escapeHTML = require('./escape_html');
+const { parse } = require('url');
 
 function htmlTag(tag, attrs, text, escape = true) {
   if (!tag) throw new TypeError('tag is required!');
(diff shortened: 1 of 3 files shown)


curbengh:
@dailyrandomphoto
any other attribute that I might've missed?

dailyrandomphoto:
> @dailyrandomphoto
> any other attribute that I might've missed?

I have tested some cases and found no issues.

curbengh:
Just found an edge case for [responsive image](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#How_do_you_create_responsive_images)

``` html
<img srcset="elva-fairy-320w.jpg 320w,
             elva-fairy-480w.jpg 480w,
             elva-fairy-800w.jpg 800w"
     sizes="(max-width: 320px) 280px,
            (max-width: 480px) 440px,
            800px"
     src="elva-fairy-800w.jpg" alt="Elva dressed as a fairy">
```

🤔 Let's see possible approaches:
1. split by space and encode each part
2. regex match string that ends with an image extension (I saw the regex pattern somewhere in hexo, can't find it), then encode it.

SukkaW:
@curbengh 

```
/<img [^>]*src=['"]([^'"]+)([^>]*>)/gi
```

This is the `` tag pattern used in `open_graph()` helper.


https://github.com/hexojs/hexo/blob/19ac152ac3fd708b1860244187442c587be73031/lib/plugins/helper/open_graph.js#L62

curbengh:
> /]*src=['"]([^'"]+)([^>]*>)/gi

That's actually is the first regex I tried, but it's not applicable here because html_tag is not parsing the a html. Essentially, the goal is not to encode the space between `elva-fairy-800w.jpg 800w`.

---

Anyway, the latest fix, while seems overcomplicating stuff, it's to preserve new lines. Frankly, I don't know if anyone actually wants it (new line) 😅 . Personally, I would just go for `handleSpaceNewLineMinify()` approach of following examples.

``` js
function handleSpace(str) {
  return encodeURI(str).replace(/(%20)/g, ' ');
}

function handleSpaceNewLine(str) {
  return encodeURI(str).replace(/(%20|%0A)/g, ' ');
}

function handleSpaceNewLineMinify(str) {
  return encodeURI(str).replace(/%20/g, ' ').replace(/,(%0A|%0D%0A)*\s*/g, ',');
}

let data = `fóo.jpg 320w,
        /foo/bár.jpeg 480w,
        default.png`

console.log(handleSpace(data))
// f%C3%B3o.jpg 320w,%0A        /foo/b%C3%A1r.jpeg 480w,%0A        default.png

console.log(handleSpaceNewLine(data))
// f%C3%B3o.jpg 320w,         /foo/b%C3%A1r.jpeg 480w,         default.png

console.log(handleSpaceNewLineMinify(data))
// f%C3%B3o.jpg 320w,/foo/b%C3%A1r.jpeg 480w,default.png
```

lib/html_tag.js
```diff
@@ -11,7 +22,8 @@ function htmlTag(tag, attrs, text, escape = true) {
   for (const i in attrs) {
     if (attrs[i] === null || typeof attrs[i] === 'undefined') result += '';
     else {
-      if (i === 'href' || i === 'src') result += ` ${i}="${encodeURL(attrs[i])}"`;
+      if (i.endsWith('href') || i.endsWith('src') || i.endsWith('url')) result += ` ${escapeHTML(i)}="${encodeURL(attrs[i])}"`;
```
SukkaW: Here are some attributes that might contain URLs:

- [`cite` attribute in `<blockquote>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote)
- [`download` attribute in `<a>` tags](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a)
- [`style` attribute in any elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/style) might have `url()` inside.

curbengh: `<style>` can a bit tricky because double-quote and slash are commonly used. what about skip escapeHTML on it, while still encode `url()`?

SukkaW: @curbengh I think that will be fine.

SukkaW: @curbengh Also, some css property like `background` `background-image` could contains URL. It is now a little more tricky. I suggest just not to encode `<style>` tag and put a notice at docs.

curbengh: Addressed in latest commits.

PR #98, opened 2019-09-20
release: 1.3.0

curbengh:
to include #92 #93 #97 #96


curbengh:
Superseded by https://github.com/hexojs/hexo-util/pull/99

PR #99, opened 2019-09-20
release: 1.3.0

curbengh:
to include #92 #93 #97 #96 #100 #101 

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hexo-util",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "description": "Utilities for Hexo.",
   "main": "lib/index",
   "scripts": {


curbengh:
Pending https://github.com/hexojs/hexo-util/pull/96

SukkaW:
After #92 #93 and #96, will we cover all edge cases that might happened? I am still worrying about the escape options might break on certain edge cases.

- Most of the URLs shouldn't be escaped.
- The html tag inside a html tag shouldn't be escaped.

Anything else?

curbengh:
>     * Most of the URLs shouldn't be escaped.

to clarify (for others reading this), URLs shouldn't be _html_ escaped.

>     * The html tag inside a html tag shouldn't be escaped.

it depends, it should be escaped when displaying as text (see https://github.com/hexojs/hexo/pull/3704), but shouldn't when used for formatting (bold, italic, etc).

e.g.

``` html
<a>How to use the <b> tag?</a>
// should be escaped
```

``` html
<a>This is <b>shocking</b>, wow</a>
// shouldn't be escaped
```

which is why there is an option to disable escaping.

curbengh:
It is possible to encode url in `<style>`, https://github.com/hexojs/hexo-util/pull/101

curbengh:
https://github.com/hexojs/hexo-util/pull/101 merged, I'm merging this.

curbengh:
https://www.npmjs.com/package/hexo-util/v/1.3.0

PR #101, opened 2019-09-22
fix(html_tag): encode url() in style tag

curbengh:
Continuation of https://github.com/hexojs/hexo-util/pull/96#discussion_r326639896

Regex is based on [`external_link.js`](https://github.com/hexojs/hexo/blob/master/lib/plugins/filter/after_post_render/external_link.js)

### DIFF
--- a/lib/html_tag.js
+++ b/lib/html_tag.js
@@ -30,6 +30,11 @@ function htmlTag(tag, attrs, text, escape = true) {
   }
 
   if (escape && text && tag !== 'style') text = escapeHTML(String(text));
+  if (text && tag === 'style') {
+    text = text.replace(/url\(['"](.*?)['"]\)/gi, (urlAttr, url) => {
+      return `url("${encodeURL(url)}")`;
+    });
+  }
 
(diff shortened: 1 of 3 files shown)


lib/html_tag.js
```diff
@@ -30,6 +30,11 @@ function htmlTag(tag, attrs, text, escape = true) {
   }
 
   if (escape && text && tag !== 'style') text = escapeHTML(String(text));
+  if (text && tag === 'style') {
+    text = text.replace(/url\(['"](.*?)['"]\)/gi, (urlAttr, url) => {
+      return `url("${encodeURL(url)}")`;
```
curbengh: this is necessary (instead of `return encodeURL(urlAttr)`) to avoid encoding single/double quote.

SukkaW:
Many css properties' value could be URLs. For example, `background` and `background-image`.

curbengh:
My understanding is that url has to be wrapped in `url("")` to be used in [`background`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-image) property,

``` css
p {
  background: url("foo.jpg");
}
```

PR #105, opened 2019-09-29
docs: Function.call() examples

curbengh:
some html tag requires a closing tag, in which case an empty string can be added as the third argument. (Edit: I moved the example to https://github.com/hexojs/hexo-util/pull/106)

`Function.call()` is based on [`full_url_for()`](https://github.com/hexojs/hexo/pull/3701/files#diff-c03d9d0d457e77e7c57b00f842473d49)

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,9 +198,6 @@ htmlTag('img', {src: 'example.png'})
 htmlTag('a', {href: 'http://hexo.io/'}, 'Hexo')
 // <a href="http://hexo.io/">Hexo</a>
 
-htmlTag('script', {src: '/foo.js'}, '')
-// <script src="foo.js"></script>
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -391,12 +394,12 @@ url_for('/css/style.css', {relative: false})
 
 ## bind(hexo)
 
-Following utilities require `bind(hexo)` / `bind(this)` to parse the user config when initializing:
+Following utilities require `bind(hexo)` / `bind(this)` / `call(this, input)` to parse the user config when initializing:
```
SukkaW:
```suggestion
Following utilities require `bind(hexo)` / `bind(this)` / `call(hexo, input)` / `call(this, input)` to parse the user config when initializing:
```

Since there is a `bind(hexo)`, the `call(hexo, input)` might need to be added as well.

PR #106, opened 2019-09-29
feat(htmlTag): support async/defer attribute

curbengh:
Intend to use this feature in https://github.com/hexojs/hexo/pull/3681 & https://github.com/hexojs/hexo/pull/3690.

### DIFF
--- a/lib/html_tag.js
+++ b/lib/html_tag.js
@@ -24,7 +24,7 @@ function htmlTag(tag, attrs, text, escape = true) {
     if (attrs[i] === null || typeof attrs[i] === 'undefined') result += '';
     else {
       if (i.match(regexUrl)) result += ` ${escapeHTML(i)}="${encodeURL(attrs[i])}"`;
-      else if (attrs[i] === true) result += ` ${escapeHTML(i)}`;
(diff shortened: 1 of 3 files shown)


lib/html_tag.js
```diff
@@ -24,6 +24,7 @@ function htmlTag(tag, attrs, text, escape = true) {
     if (attrs[i] === null || typeof attrs[i] === 'undefined') result += '';
     else {
       if (i.match(regexUrl)) result += ` ${escapeHTML(i)}="${encodeURL(attrs[i])}"`;
+      else if (attrs[i] === true) result += ` ${escapeHTML(i)}`;
```
SukkaW: https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

According to the description of `Boolean attributes`, The value of an attribute can be omitted as well if it is the same as attribute name.

curbengh: ` ${escapeHTML(i)}` only outputs ` async` not `async="true"`.

SukkaW: @curbengh 

I mean `async="async"` `disabled="disabled"` should output `async` `disabled` as well, since `Boolean attributes` not only accept Boolean value but also String value.

curbengh: Added

Issue #108, opened 2019-10-09
Feature request: support other syntax highlight libraries


curbengh:
This is a spin out of https://github.com/hexojs/hexo-util/issues/19#issuecomment-539356312.

The idea is to support different libraries, e.g. [atom's highlights](https://github.com/atom/highlights) and [prism](https://prismjs.com/), other than the current [highlight.js](https://highlightjs.org/). A major benefit is that theme dev can use/adapt from much more css. Some library may also be [faster](https://github.com/hexojs/hexo/issues/1036#issuecomment-73735029) than another. https://github.com/hexojs/hexo-util/issues/39 can be addressed by atom and prism.

---

Currently, highlight.js is tightly integrated into hexo, so adding another library is expected not to be straightforward. One possible issue is configuration, which I think should be tackled first.

So, to change library, user can specify `engine` (I will refer both `library` and `engine` interchangeably for the rest of the comment),

``` yml
highlight:
  engine: hljs|atom|prism
```

No problem, but here comes the tricky part, configuring _each_ of them. While each engine may share similar options, I think ideally they should be separated, like so:

``` yml
highlight:
  engine: ...
  hljs:
    gutter: true
  prism: ...
```

`highlight.hljs` is preferred over `highlight.highlight` for clarity. But currently the option is used to address https://github.com/hexojs/hexo-util/issues/19#issuecomment-267040668 using
``` yml
highlight:
  hljs: true
```

Retaining compatibility is possible (like https://github.com/hexojs/hexo/pull/3675), i.e. can check `highlight.hljs` is a boolean or not. If boolean, parse it as `highlight.hljs.classPrefix`.

``` yml
highlight:
  hljs:
    classPrefix: ''|'hljs-'
```

Likewise, the approach could also apply to retain compatibility with the current options:

```
highlight:
  auto_detect:
  line_number:
  tab_replace:
```

However, I suggest not to retain old options forever, they can be dropped in hexo@5 for instance, and permanently move to under `highlight.hljs`.

A side benefit of having hljs-specific option is that users familiar with the engine can use its option naming instead; hexo rename some option which confuse those users. See https://github.com/hexojs/hexo-util/issues/40.

SukkaW:
In order to retain backward compatibility for all current hexo theme, I suggest to remain current `config.highlight` there, and add a new `config.prismjs` option.

Prism related utility has been added in #168.

Related code to be changed:

[...]

PR #110, opened 2019-10-12
fix(highlight): code in blockquote

curbengh:
Currently, when inserting a codeblock inside a blockquote, e.g.

```
> ```json
> {
>   "test": 123
> }
> ```
```

results in,

> ``` json
>> {
>>  "test": 123
>> }
>```

---

Expected behavior is

>``` json
> {
>  "test": 123
> }
>```

This fixes https://github.com/hexojs/hexo/issues/2969 & https://github.com/hexojs/hexo/issues/3318.

However, this fix doesn't apply to `cmd` or `dos` codeblock because the line may start with `>`, e.g.

``` cmd
> ipconfig
```

cmd codeblock in blockquote,
```
> ```cmd
>> ipconfig
> ```
```

would result in,

> ```cmd
>> ipconfig
>>

A workaround is to remove `>` from the closing backtick,

https://pastebin.com/s94kaD4j

which should render as,

> ```cmd
> > ipconfig
> ```

or use [`{% blockquote %}`](https://hexo.io/docs/tag-plugins#Block-Quote) tag plugin.

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -61,6 +61,7 @@ function highlightUtil(str, options = {}) {
 function formatLine(line, lineno, marked, options) {
   const useHljs = options.hljs || false;
   let res = useHljs ? '' : '<span class="line';
+
   if (marked.includes(lineno)) {
     // Handle marked lines.
     res += useHljs ? `<mark>${line}</mark>` : ` marked">${line}</span>`;
@@ -132,7 +133,16 @@ function tryHighlight(str, lang) {
       html += separator + result.value;
     }
 
+    // https://github.com/hexojs/hexo/issues/2969
+    if (lang !== 'cmd' && lang !== 'dos') {
+      html = html.replace(/&gt;.*/g, (str) => {
+        str = str.replace(/^&gt;(\s)?(?!<\/span>)/, '');
+        return str;
+      }).trimRight();
+    }
+
     result.value = html;
+
     return result;
   } catch (err) {
 
--- a/test/highlight.spec.js
+++ b/test/highlight.spec.js
@@ -295,4 +295,18 @@ describe('highlight', () => {
     result.should.include('class="hljs-function"');
     validateHtmlAsync(result, done);
   });
+
+  it('codeblock inside a blockquote', (done) => {
+    const str = '> let foo = false';
+    const result = highlight(str, {gutter: false, lang: 'javascript' });
+    result.should.not.include('&gt;');
+    validateHtmlAsync(result, done);
+  });
+
+  it('codeblock inside a blockquote - exclude cmd lang', (done) => {
+    const str = '> ipconfig';
+    const result = highlight(str, {gutter: false, lang: 'cmd' });
+    result.should.include('&gt;');
+    validateHtmlAsync(result, done);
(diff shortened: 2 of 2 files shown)


lib/highlight.js
```diff
@@ -132,7 +133,16 @@ function tryHighlight(str, lang) {
       html += separator + result.value;
     }
 
+    // https://github.com/hexojs/hexo/issues/2969
+    if (lang !== 'cmd' && lang !== 'dos') {
+      html = html.replace(/&gt;.*/g, (str) => {
+        str = str.replace(/^&gt;(\s)?(?!<\/span>)/, '');
+        return str;
+      }).trimRight();
```
curbengh: this is a workaround to remove unnecessary new line from

```
>```
```

lib/highlight.js
```diff
@@ -132,7 +133,16 @@ function tryHighlight(str, lang) {
       html += separator + result.value;
     }
 
+    // https://github.com/hexojs/hexo/issues/2969
+    if (lang !== 'cmd' && lang !== 'dos') {
+      html = html.replace(/&gt;.*/g, (str) => {
+        str = str.replace(/^&gt;(\s)?(?!<\/span>)/, '');
```
curbengh: `<span>` exclusion is a workaround to be compatible with sublanguage highlight.

curbengh:
In regards to https://github.com/hexojs/hexo-util/issues/108, it may be more suitable to put this workaround in [backtick_code_block.js](https://github.com/hexojs/hexo/blob/master/lib/plugins/filter/before_post_render/backtick_code_block.js), but currently I'm not sure how well prism/atom handle blockquote.

If prism/atom cannot handle blockquote, then I'll move this workaround to hexo.

seaoak:
I think a workaround in the hexo ([backtick_code_block.js](https://github.com/hexojs/hexo/blob/master/lib/plugins/filter/before_post_render/backtick_code_block.js)) might be better
because the module is responsible to the extraction of code block (and blockquote).

I found examples of markdown notation in the GitHub Help:
* https://github.github.com/gfm/#fenced-code-blocks
* https://github.github.com/gfm/#block-quotes

There are many examples which current `backtick_code_block.js` can not process correctly:
- [Example 98](https://github.github.com/gfm/#example-98)
- [Example 101](https://github.github.com/gfm/#example-101)
- [Example 102](https://github.github.com/gfm/#example-102)
- [Example 103](https://github.github.com/gfm/#example-103)
- [Example 215](https://github.github.com/gfm/#example-215)
- [Example 228](https://github.github.com/gfm/#example-228)
- [Example 229](https://github.github.com/gfm/#example-229)

As shown in [Example 229](https://github.github.com/gfm/#example-229),
the question "how many '>' at the head of line should be removed"
is only answerable by Markdown parser.
A code-highlighting module is not responsible to this question.
(And also it is not affected by "lang" parameter of the code block.)

I'm trying to improve `backtick_code_block.js`,
but it is almost like developing small set of "marked".
I started from check the behavior of "marked" about the examples....

curbengh:
Superseded by https://github.com/hexojs/hexo/pull/3765.

PR #112, opened 2019-10-19
Make consistency in output of "highlight()" for "wrap: false" (fix #111)

### LINKED ISSUE #111: `highlight.js` generates no HTML class name when "wrap: false" and "hljs: false"
This issue is the behavior of `lib/highlight.js`.

In case of `wrap: false`:
  If `hljs: true`, an HTML class name `hljs` appears in the output.
  If `hljs: false`, no HTML class name appears in the output.

It is a lack of consistency.

I think an HTML class name `highlight` should appear as well as when `wrap: true`.

And also, the output for `hljs: false` lacks a PRE element and a CODE element.
That is, no HTML element appears around the output.
It is a lack of consistency, too.

## Expected behavior

The output of `util.highlight('alert("foobar")', {wrap: false})` is:

```html
<pre><code class="highlight">alert(&quot;foobar&quot;)</code></pre>
```

## Actual behavior

```html
alert(&quot;foobar&quot;)
```

## How to reproduce?

```
$ node
> const util = require('.')
undefined
> util.highlight('alert("foobar")', {wrap: false})
'alert(&quot;foobar&quot;)'
> util.highlight('alert("foobar")', {wrap: false, hljs: true})
'<pre><code class="hljs undefined">alert(&quot;foobar&quot;)</code></pre>'
>
```

## Environment & Settings

```
$ node -v
v8.16.2

$ npm -v
6.4.1

$ git log --oneline --decorate -1
b292332 (HEAD -> master, upstream/master, origin/master, origin/HEAD) Merge pull request #109 from curbengh/toc
```

seaoak:
Fix #111

This patch is fix the behavior of `highlight()` with a option `wrap: false`.

If `hljs: false`, the output of `highlight()` should include a HTML class name `highlight`.
It is consistent with the case of `hljs: true`.

And also, even if `hljs: false`, the output of `highlight()` should be include a PRE element and a CODE element.
It is consistent with the case of `hljs: true`, too.

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -21,13 +21,16 @@ function highlightUtil(str, options = {}) {
   hljs.configure({ classPrefix: useHljs ? 'hljs-' : ''});
 
   const data = highlight(str, options);
+  const lang = options.lang || data.language || '';
+  const classNames = (useHljs ? 'hljs' : 'highlight') + (lang ? ` ${lang}` : '');
 
   if (useHljs && !gutter) wrap = false;
+  if (gutter && !wrap) wrap = true; // arbitrate conflict ("gutter:true" takes priority over "wrap:false")
 
-  const before = useHljs ? `<pre><code class="hljs ${options.lang}">` : '<pre>';
+  const before = useHljs ? `<pre><code class="${classNames}">` : '<pre>';
   const after = useHljs ? '</code></pre>' : '</pre>';
 
-  if (!wrap) return useHljs ? before + data.value + after : data.value;
+  if (!wrap) return `<pre><code class="${classNames}">${data.value}</code></pre>`;
 
   const lines = data.value.split('\n');
   let numbers = '';
--- a/test/highlight.spec.js
+++ b/test/highlight.spec.js
@@ -85,9 +85,51 @@ describe('highlight', () => {
     validateHtmlAsync(result, done);
   });
 
-  it('wrap: false', done => {
-    const result = highlight(testString, {wrap: false});
-    result.should.eql(entities.encode(testString));
+  it('gutter: true, but "wrap: false" (conflict)', done => {
+    const result = highlight(testString, {gutter: true, wrap: false});
+    assertResult(result, gutter(1, 4), code(testString));
+    validateHtmlAsync(result, done);
+  });
+
+  it('wrap: false (without hljs, without lang)', done => {
+    const result = highlight(testString, {gutter: false, wrap: false});
+    result.should.eql([
+      '<pre><code class="highlight plain">',
+      entities.encode(testString),
+      '</code></pre>'
+    ].join(''));
+    validateHtmlAsync(result, done);
+  });
+
+  it('wrap: false (with hljs, without lang)', done => {
+    const result = highlight(testString, {gutter: false, wrap: false, hljs: true});
+    result.should.eql([
+      '<pre><code class="hljs plain">',
+      entities.encode(testString),
+      '</code></pre>'
+    ].join(''));
+    validateHtmlAsync(result, done);
+  });
+
+  it('wrap: false (without hljs, with lang)', done => {
+    const result = highlight(testString, {gutter: false, wrap: false, lang: 'json'});
+    hljs.configure({classPrefix: ''});
+    result.should.eql([
+      '<pre><code class="highlight json">',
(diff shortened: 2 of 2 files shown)


seaoak:
This PR will be used to introduce `wrap` option in `highlight` section of Hexo's `_config.yml`.

lib/highlight.js
```diff
@@ -21,13 +21,15 @@ function highlightUtil(str, options = {}) {
   hljs.configure({ classPrefix: useHljs ? 'hljs-' : ''});
 
   const data = highlight(str, options);
+  const lang = options.lang || ''; // Maybe "data.language" should be included
```
curbengh: My understanding is that `data.language` would default to `plain` value if no lang is provided. In current behavior `wrap = true`, `plain` does get appended to class name -> `"highlight plain"`.

So, I agree with `options.lang || data.language || '';`.

seaoak: One thing I worry about is that the default behavior of Hexo with `hljs: true` will be changed.

Under the code `options.lang || data.language || ''`,
when `hljs:true` is specified with neither `lang` option nor `gutter` option,
the output of `highlight()` will be `<pre><code class="hljs plain">...`.

On the other hand, the current implementation says `<pre><code class="hljs undefined">...`.
That is, a class name `plain` is not added (adding `undefined` is a tiny bug).

Under the code `options.lang || ''` (without `data.language`),
the default output of `highlight()` will be `<pre><code class="hljs">...`.
That is, no class name is added.

So including `data.language` causes Hexo changing the default behavior with `hljs: true, line_number: false`.
Is it OK?

curbengh: > the current implementation says `<pre><code class="hljs undefined">`

which implementation are you referring to? this PR with `options.lang || data.language || ''`?

---

The main reason of having "highlight plain"  is so that "hljs: false, wrap: false" can be compatible with current theme as much as possible. I just worry that some theme might use `"plain"` class.

> So including data.language causes Hexo changing the default behavior with `hljs: true, line_number: false`.

I don't foresee any breaking change for adding a new class (i.e. "plain"), even when it's not used in the newer syntax.

---

In summary, without `data.language`, "plain" class is removed when "hljs: false, wrap: false" which is a possible breaking change. Having an unused "plain" class in "hljs: true" is not a breaking change.

seaoak: @curbengh Thank you for explaining.  I didn't understand enough.
Now I understand that `data.language` is necessary to keep compatibility with current themes and adding a class `plain` will not break current themes.

I'll update the code soon.

seaoak:
I updated the code of this PR (reflect @curbengh 's comments and add two test cases).
Could you review this again?

curbengh:
Since `wrap: false` is not compatible with `gutter: true`, shall we give priority to "gutter" (i.e. enable "wrap" when "gutter")? My preference is on "gutter".

Definitely we'll have to mention in the docs about the compatibility (after `wrap` option is introduced in hexo).

seaoak:
I agree with giving the priority (`gutter` option is superior to `wrap` option).

I will update this PR to arbitrate the conflict with them.

lib/highlight.js
```diff
@@ -21,13 +21,16 @@ function highlightUtil(str, options = {}) {
   hljs.configure({ classPrefix: useHljs ? 'hljs-' : ''});
 
   const data = highlight(str, options);
+  const lang = options.lang || data.language || '';
+  const classNames = (useHljs ? 'hljs' : 'highlight') + (lang ? ` ${lang}` : '');
 
   if (useHljs && !gutter) wrap = false;
+  if (gutter && !wrap) wrap = true; // arbitrate conflict ("gutter:true" takes priority over "wrap:false")
```
seaoak: I update this PR.  I added a code to arbitrate conflict with `gutter:` and `wrap:`.
I think this arbitration will be meaningful for adding `wrap:` option into Hexo's `_config.yml`.

@curbengh Could you review again?

curbengh: Somehow hexo binary is broken when installing from this PR, can you try rebasing?

curbengh: Can you add a test case for `"gutter: true, wrap false"`?

seaoak:
@curbengh I'm sorry it was an elementary mistake.
I updated this PR (adding a test case).

PR #114, opened 2019-10-22
refactor: whatwg url api

curbengh:
legacy URL api has been deprecated in Node v11; while the API is still available in Node 12, `eslint-plugin-node` considers it as a deprecated api, thus will emit error if detected.

WHATWG url is actually introduced back in Node v7, thus contrary to my previous PR (https://github.com/hexojs/eslint-config-hexo/pull/19), it is supported in Node 8. The only thing is that Node 8 requires `const {URL} = require('url')`, whereas newer Node doesn't.

Another noticeable difference is that WHATWG remove the port number if it corresponds to the protocol (e.g. `http`:`80`, `https`:`443`), which explains the change of port number in the unit test.

`decode_url` is not refactored to the newer API for now due to a bug in Node 8,

``` js
const {format} = require('url')
const bob = new URL('http://user:pass@foo.com:8080/')
console.log(format(bob, { unicode: true }))
// http://user:pass@/
// Domain is removed if port number exists
```

Once we drop Node 8, not only we can update `decode_url`, we can also remove `const {URL} = require('url')` line. It's ok to retain it, but it's not necessary.

### DIFF
--- a/lib/decode_url.js
+++ b/lib/decode_url.js
@@ -1,5 +1,9 @@
 'use strict';
 
+// To be refactored to WHATWG URL API
+// after support for Node 8 is dropped.
+// url.format(WHATWG URL object) in Node 8 doesn't support port number
+
 const { parse, format } = require('url');
 const { toUnicode } = require('./punycode');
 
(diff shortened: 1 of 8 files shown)


curbengh:
Added a test for 4964d808a87a7275c4563ff0b1352bfb61a4b818.

PR #116, opened 2019-10-27
Leave IDN intact in permalink

tomap:
Because IDN only contain unicode, they do not need to be punycode encoded.
(I bought one, I want the nice version, not the ugly one)
See https://nodejs.org/docs/latest/api/url.html#url_url_format_url_options

### DIFF
--- a/lib/encode_url.js
+++ b/lib/encode_url.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { URL } = require('url');
+const { URL, format } = require('url');
 
 const urlObj = (str) => {
   try {
@@ -24,7 +24,8 @@ const encodeURL = (str) => {
     if (parsed.origin === 'null') return str;
 
     parsed.search = encodeURI(safeDecodeURI(parsed.search));
-    return parsed.toString();
+    // preserve host (international domain name) as is.
+    return format(parsed, { unicode: true});
   }
 
   return encodeURI(safeDecodeURI(str));
--- a/lib/encode_url.js
+++ b/lib/encode_url.js
@@ -1,6 +1,7 @@
 'use strict';
 
-const { URL, format } = require('url');
+const { toUnicode } = require('./punycode');
+const { URL } = require('url');
 
 const urlObj = (str) => {
   try {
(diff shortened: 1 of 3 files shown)


SukkaW:
Maybe we should implement two `encode_url()` method for different usages.

tomap:
I'm not sure there is a usage where IDN needs to be punycode encoded.

lib/encode_url.js
```diff
@@ -24,7 +24,8 @@ const encodeURL = (str) => {
     if (parsed.origin === 'null') return str;
 
     parsed.search = encodeURI(safeDecodeURI(parsed.search));
-    return parsed.toString();
+    // preserve host (international domain name) as is.
+    return format(parsed, { unicode: true});
```
curbengh: url.format() doesn't support port number in Node 8, see https://github.com/hexojs/hexo-util/pull/114.

curbengh:
@tomap 

what is your use case?

``` html
<a href="punycoded-domain">text</a>
```

Punycoded-domain plus the percent-encoded path are decoded when mouse-over.

---

``` html
<a href="punycoded-domain">url as text</a>
```

Usually user doesn't want to display encoded url (punycode/percent-encoding) as text, so the url can be decoded using [`decodeURL()`](https://github.com/hexojs/hexo-util#decodeurlstr). decodeURL decodes both punycode and percent-encoding.

lib/encode_url.js
curbengh: A workaround is to use punycode api

``` js
const { toUnicode } = require('./punycode');
const url = parsed.toString().replace(parsed.hostname, toUnicode(parsed.hostname);
```

curbengh: readme also needs to be updated.

curbengh:
> I'm not sure there is a usage where IDN needs to be punycode encoded.

When permalink is encoded by default, it breaks `encodeURI()` (permalink is encoded twice).

If a plugin wants to remain compatible with <3.9, they can use `encodeURI(decodeURI())` or `encodeURL()` instead.

If the IDN is not punycoded, `encodeURI(decodeURI())` will percent-encode the domain.

But of course, as we learned from hexo-generator-feed & hexo-generator-sitemap, `encodeURI()`/`encodeURI(decodeURI())` should be avoided and use `encodeURL()` instead.

Note that I still +1 this PR.

lib/encode_url.js
tomap: node 8 EOL is arround the corner, and will probably imply that this package, as well as many other and also hexo have a major version bump

curbengh: we can use the workaround temporarily until node 8 EOL. I don't think this is a breaking change, I do agree IDN users would prefer not to punycode.

curbengh: Added the workaround. another change introduced by this PR is that IDN whether it's in unicode or punycoded is always encoded to unicode.

``` js
encodeURL(http://xn--br-mia.com/baz')
// http://bár.com/baz
```

PR #117, opened 2019-10-28
refactor(decodeURL): update to whatwg api

curbengh:
Continuation of #114 

Node 8's `url.format()` doesn't support port number, so a workaround is used instead.


curbengh:
Ready for review.

PR #119, opened 2019-10-28
feat: bring up isExternalLink

SukkaW:
Returns if a given url is external link relative to `config.url` and `config.exclude`.

Can be utilized in Hexo's `external_link` filter and `hexo-filter-nofollow` plugin.


test/isExternalLink.spec.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+describe('isExternalLink', () => {
+  const ctx = {
+    config: {
+      url: 'https://example.com',
+      external_link: {
+        exclude: ''
```
curbengh: it is possible that `exclude` can be [`undefined`](https://github.com/hexojs/hexo-starter/blob/80b8ec43ddd4851ffe7b4757c85d779e4c922b46/_config.yml#L40). can you add a test case for that?

SukkaW: Since `external_link.exclude` is empty by [default](https://github.com/hexojs/hexo-starter/blob/80b8ec43ddd4851ffe7b4757c85d779e4c922b46/_config.yml#L40), then the default test case should use undefined then.

curbengh: I think it's ok to test for empty string `''`. Just to be safe, suggest to test for `undefined` and `null`.

btw, I submitted a PR to default `exclude` to `''`. https://github.com/hexojs/hexo-starter/pull/26 , but it's still possible the value to be `undefined`.

curbengh:
``` js
const exclude = Array.isArray(config.external_link.exclude) ? config.external_link.exclude
  : [config.external_link.exclude];

  if (exclude && exclude.length) {
    for (const i of exclude) {
      if (host === i) return false;
    }
  }
```

I tested `undefined` value. `exclude` ends up to be `[undefined]`, so the for loop did execute. no error though, to my surprise.

anyway, `[undefined]` looks a bit grotesque, which is why I suggested to test `undefined`.

SukkaW: I have already update test case to 

```js
  const ctx = {
    config: {
      url: 'https://example.com',
      external_link: {}
    }
  };
```

to test undefined,

SukkaW:
@curbengh I might change filename to `is_external,js`

curbengh:
> @curbengh I might change filename to `is_external_link.js`

please do.

SukkaW:
@curbengh Filename changed in https://github.com/hexojs/hexo-util/pull/119/commits/a4a2873f5677d0c962ac933338157b35f078dc97

PR #123, opened 2019-11-08
fix(isExternal): url parameter can be an internal link

curbengh:
I just noticed a silent issue.

``` js
const data = urlObj(url);
const host = data.hostname;
```

`data` will be a string when `url` is an internal link, resulting in `host` being assigned as `undefined`. when it is "undefined", it won't be used, thanks to

``` js
if (!sitehost || typeof data === 'string') return false;
```

While current code won't cause any crash, it's a bit unsafe to leave it "undefined" and also could lead to incorrect assumption that `data` is always a URL object.

### DIFF
--- a/lib/is_external_link.js
+++ b/lib/is_external_link.js
@@ -21,7 +21,7 @@ function isExternalLink(url) {
   const exclude = Array.isArray(config.external_link.exclude) ? config.external_link.exclude
     : [config.external_link.exclude];
   const data = urlObj(url);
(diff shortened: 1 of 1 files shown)


SukkaW:
Might have a look at https://github.com/hexojs/hexo/pull/3839 ? In this PR I bring up a poc on performance. In that PR I also fix this issue.

curbengh:
https://github.com/hexojs/hexo-util/pull/124

PR #124, opened 2019-11-08
refactor(isExternal): avoid undefined & improve performance

SukkaW:
The PR fixes #123 with a simpler way. This PR also includes the changes from https://github.com/hexojs/hexo/pull/3839 , which is a poc of performance improvements.

### DIFF
--- a/lib/is_external_link.js
+++ b/lib/is_external_link.js
@@ -20,14 +20,18 @@ function isExternalLink(input) {
   if (data.origin === 'null') return false;
 
   const host = data.hostname;
-  const exclude = Array.isArray(config.external_link.exclude) ? config.external_link.exclude
-    : [config.external_link.exclude];
 
-  if (exclude && exclude.length) {
-    for (const i of exclude) {
-      if (host === i) return false;
+  if (config.external_link && config.external_link.exclude) {
+    const exclude = Array.isArray(config.external_link.exclude) ? config.external_link.exclude
+      : [config.external_link.exclude];
+
+    if (exclude && exclude.length) {
+      for (const i of exclude) {
+        if (host === i) return false;
+      }
     }
   }
+
 
   if (host !== sitehost) return true;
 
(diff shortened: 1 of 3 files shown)


lib/is_external_link.js
```diff
@@ -18,17 +10,19 @@ const urlObj = (str) => {
 
 function isExternalLink(url) {
   const { config } = this;
-  const exclude = Array.isArray(config.external_link.exclude) ? config.external_link.exclude
-    : [config.external_link.exclude];
-  const data = urlObj(url);
-  const host = data.hostname;
-  const sitehost = typeof urlObj(config.url) === 'object' ? urlObj(config.url).hostname : config.url;
+  const sitehost = parse(config.url).hostname || config.url;
+  // Pass a base to new URL
```
curbengh: I would say the purpose is `// Workaround for relative url`.

lib/is_external_link.js
```diff
@@ -18,17 +10,19 @@ const urlObj = (str) => {
...
+  const data = new URL(url, `http://${sitehost}`);
 
   if (!sitehost || typeof data === 'string') return false;
```
curbengh: with L15, `typeof data` is no longer required since it will always be an "object".

lib/is_external_link.js
```diff
@@ -18,17 +10,19 @@ const urlObj = (str) => {
 
 function isExternalLink(url) {
```
curbengh: suggest to rename to `path` or `input` to not confuse with `new URL()`.

lib/is_external_link.js
SukkaW: Updated to `// handle relative url` since there is a `handle mailto: javascript: vbscript: and so on` below.

lib/is_external_link.js
SukkaW: Removed.

lib/is_external_link.js
SukkaW: `input` might be a good idea.

curbengh:
It's fine to use `isExternalLink(url)` in documentation.

SukkaW:
@curbengh So should we should throw a TypeError if `config.url` is null or undefined?

Issue #122, opened 2019-11-07
url_for helper doesn't work for mailto link


lookis:
https://github.com/hexojs/hexo-util/blob/master/lib/url_for.js#L28-L31

it will return origin path when met external link like http://example.com, but will not return origin path for mailto: someone@example.com. and lead to an issue when relative_link was set.

url_for('/category') => /relative_link/category <= OK!
url_for('http://example.com') => http://example.com <= OK!
url_for('mailto:someone@example.com') => /relative_link/mailto:someone@example.com <= Ops!

SukkaW:
There is no better way to fix this. We might need to hardcode some whitelist protocol inside hexo-util, just as one of contributor of nodejs gives: https://github.com/nodejs/node/pull/28482/

https://github.com/nodejs/node/blob/49a5fa0d8b15cc50df833848bfc01d2260930dfd/lib/internal/url.js#L340-L348

SukkaW:
@curbengh Any idea?

In `isExternalLink()` we use `new URL(path)).origin === 'null'` to consider `mailto:` `javascript:` links as an internal link. But when at `url_for` and `full_url_for`, it requires to consider them as an external link.

Currently I only come up with the idea of whitelist protocols.

curbengh:
The [workaround](https://github.com/hexojs/hexo-util/blob/master/lib/url_for.js#L28-L31) is necessary to support path with [a semicolon](https://github.com/hexojs/hexo-util/blob/master/test/url_for.spec.js#L76-L81).

However, since `encodeURL()` never support path with semicolon, I'm ok dropping it from `url_for()` and `full_url_for()`.

I'm assuming `mailto:` is more common than path with semicolon.

curbengh:
When I created https://github.com/hexojs/hexo-util/pull/114, I assumed `url_for()` would only be used for relative url, not data url (like `mailto:`) nor absolute url, which is why I made a leeway to support path with semicolon.

In relation to https://github.com/theme-next/hexo-theme-next/issues/1260, I wonder why  `url_for(mailto:)` exists (my apology, I don't understand `社交栏` and `检查单`). I assume the theme dev also share my assumption.

Btw, this issue is still going to [be fixed](https://github.com/hexojs/hexo-util/pull/130).

lookis:
thx @curbengh 
社交栏 means social account such as facebook or twitter account
检查单 is a check list for user to submit an issue

and I am not the author of theme-next, maybe they wrap every link with url_for :(

[...]

PR #125, opened 2019-11-09
perf: utilize legacy url.parse() in url_for() & full_url_for()

curbengh:
Continuation of https://github.com/hexojs/hexo-util/pull/124

Ref: https://github.com/hexojs/hexo/issues/3846

### DIFF
--- a/lib/full_url_for.js
+++ b/lib/full_url_for.js
@@ -1,25 +1,16 @@
 'use strict';
 
-const { URL } = require('url');
+const { parse, URL } = require('url');
 const encodeURL = require('./encode_url');
-
-const urlObj = (str) => {
-  try {
-    return new URL(str);
-  } catch (err) {
-    return str;
-  }
-};
 
 function fullUrlForHelper(path = '/') {
   if (path.startsWith('//')) return path;
   const { config } = this;
-  const data = urlObj(path);
+  const sitehost = parse(config.url).hostname || config.url;
+  const data = new URL(path, `http://${sitehost}`);
 
-  // Exit if this is an external path
-  if (typeof data === 'object') {
-    if (data.origin !== 'null') return path;
-  }
+  // Exit if this is an external link
+  if (data.hostname && data.hostname !== sitehost) return path;
 
   path = encodeURL(config.url + `/${path}`.replace(/\/{2,}/g, '/'));
   return path;
--- a/lib/full_url_for.js
+++ b/lib/full_url_for.js
@@ -2,7 +2,6 @@
 
 const { parse, URL } = require('url');
 const encodeURL = require('./encode_url');
-const isExternalLink = require('./is_external_link');
 
 function fullUrlForHelper(path = '/') {
   if (path.startsWith('//')) return path;
@@ -12,7 +11,7 @@ function fullUrlForHelper(path = '/') {
   const data = new URL(path, `http://${sitehost}`);
 
   // Exit if this is an external link
-  if (isExternalLink.call(this, path)) return path;
+  if (data.origin !== 'null' && data.hostname !== sitehost) return path;
 
   path = encodeURL(config.url + `/${path}`.replace(/\/{2,}/g, '/'));
   return path;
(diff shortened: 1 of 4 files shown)


curbengh:
``` js
if (data.hostname && data.hostname !== sitehost) return path;
```

`if (data.hostname)` seems redundant, but it is necessary because `new URL('path:withSemicolon', `http://${siteHost}).hostname` results in empty string `''`. Adding that condition is avoid early exit and force it to process path with semicolon.

SukkaW:
I notice `config.url` is added to the test case. So is it means after this PR we might not be able to handle `undefined` case? For `isExternalLink()`, it will `return false` when `(!config.url)` is true.

curbengh:
Good point, now that I think about it, it's more accurate to throw an error if `config.url` is undefined/null, since `full_url_for()` and `isExternal()` wouldn't work.

test/full_url_for.spec.js
```diff
@@ -2,7 +2,12 @@
 
 describe('full_url_for', () => {
   const ctx = {
-    config: {}
+    config: {
+      url: 'http://example.com',
+      external_link: {
+        exclude: []
+      }
```
curbengh: Need condition checking in `isExternalLink()` to skip `exclude` checking, if `config.external_link` doesn't exist.

lib/full_url_for.js
```diff
@@ -1,25 +1,18 @@
...
-  const data = urlObj(path);
+  const sitehost = parse(config.url).hostname || config.url;
+  const data = new URL(path, `http://${sitehost}`);
 
-  // Exit if this is an external path
-  if (typeof data === 'object') {
-    if (data.origin !== 'null') return path;
-  }
+  // Exit if this is an external link
+  if (isExternalLink.call(this, path)) return path;
```
SukkaW: Emmmm. I don't think it is a good idea to call isExternalLink here.
url_for is used to complete the relative path while full_url_for not only complete relative path but also add domain prefix to make a absolute url.

test/full_url_for.spec.js
SukkaW: Fixed in https://github.com/hexojs/hexo-util/pull/124/commits/34088245fc4f576321546a7dab4cea10eb8ffd74 (#124)

lib/full_url_for.js
SukkaW: `isExternalLink()` takes `external_link.exclude` into consideration while `url_for` and `full_url_for` definitely don't need. So it becomes quite a clear. We can simply use `if (data.hostname !== sitehost) return path;` and it will be all fine.

lib/url_for.js
```diff
@@ -1,35 +1,27 @@
...
   }
 
   const { config } = this;
   const { root } = config;
-  const data = urlObj(path);
+  const sitehost = parse(config.url).hostname || config.url;
+  const data = new URL(path, `http://${sitehost}`);
+
+  // Exit if this is an external link
+  if (isExternalLink.call(this, path)) return path;
```
SukkaW: Same here.

```suggestion
  if (data.origin !== 'null' || data.hostname !== sitehost) return path;
```

lib/full_url_for.js
SukkaW:
```suggestion
  if (data.origin !== 'null' || data.hostname !== sitehost) return path;
```

https://github.com/hexojs/hexo-util/blob/34088245fc4f576321546a7dab4cea10eb8ffd74/lib/is_external_link.js#L20
https://github.com/hexojs/hexo-util/blob/34088245fc4f576321546a7dab4cea10eb8ffd74/lib/is_external_link.js#L36

The main logic of `isExternalLink()` without `config.external_link.exclude`

lib/url_for.js
```diff
@@ -16,7 +15,7 @@ function urlForHelper(path = '/', options) {
   const data = new URL(path, `http://${sitehost}`);
 
   // Exit if this is an external link
-  if (isExternalLink.call(this, path)) return path;
+  if (data.origin !== 'null' && data.hostname !== sitehost) return path;
```
SukkaW: May accidentally fix #122?

Maybe add a test case?

curbengh: #122 requires the reverse, exit if `data.origin === 'null'.

I'll fix #122 in a separate PR (after this), #122 requires dropping support for [path with semicolon](https://github.com/hexojs/hexo-util/blob/master/test/url_for.spec.js#L76-L81).

PR #129, opened 2019-11-10
refactor: relpace node-html-entities with escapeHTML()

### LINKED ISSUE #127: Drop node-html-entities
https://github.com/hexojs/hexo-util/blob/42b7393b9c591d3874e0c687b4dba5a17512121a/lib/highlight.js#L79

hexo-utils built-in highlight is using `node-html-entities`, a lib that hasn't been active for last 4 years.

I might take a look at this and try to drop this dependency.

SukkaW:
Fix #127. Use hexo-util's built-in `escapeHTML()` to replace `node-html-entities`, while retains `node-html-entities` in unit-test to make sure the behaviors of highlight remain unchanged after this PR.

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -1,9 +1,9 @@
 'use strict';
 
 const hljs = require('highlight.js');
-const Entities = require('html-entities').XmlEntities;
-const entities = new Entities();
 const alias = require('../highlight_alias.json');
+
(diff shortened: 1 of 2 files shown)


SukkaW:
Rebased to remove accidentally including `package-lock.json`.

PR #130, opened 2019-11-10
fix(url_for & full_url_for): ignore data url

### LINKED ISSUE #122: url_for helper doesn't work for mailto link

curbengh:
Fixes #122 

BREAKING CHANGE: drop support for pathname with semicolon

cc @lookis

### DIFF
--- a/lib/full_url_for.js
+++ b/lib/full_url_for.js
@@ -10,8 +10,8 @@ function fullUrlForHelper(path = '/') {
   const sitehost = parse(config.url).hostname || config.url;
   const data = new URL(path, `http://${sitehost}`);
 
-  // Exit if this is an external link
-  if (data.origin !== 'null' && data.hostname !== sitehost) return path;
+  // Exit if input is a data url or external link
+  if (data.origin === 'null' || data.hostname !== sitehost) return path;
 
   path = encodeURL(config.url + `/${path}`.replace(/\/{2,}/g, '/'));
   return path;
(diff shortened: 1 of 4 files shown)


lib/full_url_for.js
```diff
@@ -10,8 +10,8 @@ function fullUrlForHelper(path = '/') {
   const sitehost = parse(config.url).hostname || config.url;
   const data = new URL(path, `http://${sitehost}`);
 
-  // Exit if this is an external link
-  if (data.origin !== 'null' && data.hostname !== sitehost) return path;
+  // Exit if input is a data url or external link
+  if (data.origin === 'null' || data.hostname !== sitehost) return path;
```
SukkaW:
```suggestion
  if (data.hostname !== sitehost || data.origin === 'null') return path;
```

IMHO, external link should be more common than `mailto:`, `javascript:` url. If we change the order, we might increase performance by just a little bit.

curbengh: updated

PR #134, opened 2019-11-17
feat: apply pretty_url to url_for & full_url_for

SukkaW:
I have noticed `url_for` & `full_url_for` will not remove `index.html` while `trailing_index` is disabled. This PR is addressed to make the `trailing_index` behavior more consistent. 

I haven't apply pretty_url options to `relative_url()` yet because relative_url doesn't require `bind(hexo)` and can not read `hexo.config` directly. After we finish `relative_url()` I will make this PR "Ready to Review".

### DIFF
--- a/lib/full_url_for.js
+++ b/lib/full_url_for.js
@@ -14,6 +14,13 @@ function fullUrlForHelper(path = '/') {
   if (data.hostname !== sitehost || data.origin === 'null') return path;
 
   path = encodeURL(config.url + `/${path}`.replace(/\/{2,}/g, '/'));
+
+  const { trailing_index } = Object.assign({
+    trailing_index: true
+  }, config.pretty_url);
+
+  if (!trailing_index) path = path.replace(/index\.html$/, '');
+
   return path;
 }
 
--- a/lib/url_for.js
+++ b/lib/url_for.js
@@ -5,9 +5,7 @@ const encodeURL = require('./encode_url');
 const relative_url = require('./relative_url');
 
 function urlForHelper(path = '/', options) {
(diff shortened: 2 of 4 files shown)


curbengh:
LGTM too.

curbengh:
how about adding an optional boolean parameter to `relative_url()`?

`relative_url(from, to, true)`, then remove trailing `index.html` if false.

SukkaW:
@curbengh `pretty_url` will have `trailing_index`, `trailing_html`, maybe even more option in the future. A simple boolean parameter just won't help.

curbengh:
> `pretty_url` will have `trailing_index`, `trailing_html`, maybe even more options in the future.

what about [`codeblock()`](https://github.com/hexojs/hexo/blob/master/lib/plugins/tag/code.js) tag plugin approach? e.g. `relative_url(from, to, { trailing_index: true, trailing_html: true })`.

Another idea I just thought of is that making `bind(hexo)`/`call()` optional for `url_for()`/`full_url_for()`/others (including `isExternalLink()`, where those functions can be used as `url_for(input, siteConfig)`. Those functions can check existence of `this.config.url` and `siteConfig.url` and respond appropriately.

This is useful where `this` object doesn't refer to `hexo` (like in [hexo-renderer-marked](https://github.com/hexojs/hexo-renderer-marked/blob/f3efcd0c66f92d4d01a28a5f25e5b2d2358976be/lib/renderer.js#L76)) or the `hexo` object can't be accessed directly.

SukkaW:
I think pass a `config` as the last parameter could be a good idea. We first check if that parameter is given, if doesn't, then get config from `this`.

PR #141, opened 2019-11-22
feat(deepMerge): bring up

SukkaW:
Continuation of https://github.com/hexojs/hexo/pull/3880

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,14 +114,24 @@ decodeURI(format(new URL('http://xn--br-mia.com.com/b%C3%A1r'), {unicode: true})
 
 ### deepMerge(target, source)
 
-Merges the enumerable properties of two objects deeply. Only `target` object will be changed.
+Merges the enumerable properties of two objects deeply while neither `target` or `source` is modified.
 
 ``` js
+// Merge deeply
 const obj1 = {a: {b: 1, c: 1, d: {e: 1, f: 1}}};
 const obj2 = {a: {b: 2, d: {f: 'f'} }};
 
 deepMerge(obj1, obj2);
 // {a: {b: 2, c: 1, d: {e: 1, f: 'f'} }}
+```
+
+``` js
+// Objects will be combined at the same index in the two arrays, just as lodash.merge
+const obj1 = { 'a': [{ 'b': 2 }, { 'd': 4 }] };
+const obj2 = { 'a': [{ 'c': 3 }, { 'e': 5 }] };
+
+deepMerge(obj1, obj2);
+// { 'a': [{ 'b': 2, 'c': 3 }, { 'd': 4, 'e': 5 }] };
 ```
 
 ### encodeURL(str)
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ decodeURI(format(new URL('http://xn--br-mia.com.com/b%C3%A1r'), {unicode: true})
 
 ### deepMerge(target, source)
 
-Merges the enumerable properties of two objects deeply.
+Merges the enumerable properties of two objects deeply. Only `target` object will be changed.
 
 ``` js
 const obj1 = {a: {b: 1, c: 1, d: {e: 1, f: 1}}};
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ decodeURI(format(new URL('http://xn--br-mia.com.com/b%C3%A1r'), {unicode: true})
 
 ### deepMerge(target, source)
 
-Merges the enumerable properties of two objects deeply while neither `target` or `source` is modified.
+Merges the enumerable properties of two objects deeply. `target` and `source` remain untouched.
 
 ``` js
 // Merge deeply
@@ -126,7 +126,7 @@ deepMerge(obj1, obj2);
 ```
 
 ``` js
-// Objects will be combined at the same index in the two arrays, just as lodash.merge
+// Arrays will be combined in the same property, similar to lodash.merge
 const obj1 = { 'a': [{ 'b': 2 }, { 'd': 4 }] };
 const obj2 = { 'a': [{ 'c': 3 }, { 'e': 5 }] };
 
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Utilities for [Hexo].
 - [camelCaseKeys](#camelcasekeysobj-options)
 - [createSha1Hash](#createsha1hash)
 - [decodeURL](#decodeurlstr)
+- [deepMerge](#deepmergetarget-source)
 - [encodeURL](#encodeurlstr)
 - [escapeDiacritic](#escapediacriticstr)
 - [escapeHTML](#escapehtmlstr)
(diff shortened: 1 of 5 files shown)


curbengh:
I'm curious about the performance difference (if any) between native and lodash.merge.

SukkaW:
@curbengh From benchmark of https://github.com/hexojs/hexo/pull/3880 it seems that it is more or less the same as the master branch.

SukkaW:
| | Node.js 8 | Node.js 10 | Node.js 12 | Node.js 13 |
|:---:|:---:|:---:|:---:|:---:|
| Current master branch of Hexo | 38.0s | 39.8s | 81.7s | 85.3s |
| PR #3835 | 40.0s | 35.0s | 87.9s | 83.5s |

lib/deep_merge.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+function deepMerge(target, source) {
+  target = Object.assign({}, target);
+  source = Object.assign({}, source);
```
curbengh: need to be removed, otherwise

``` js
const obj1 = {a: {b: 1, c: 1, d: {e: 1, f: 1}}};
const obj2 = {a: {b: 2, d: {f: 'f'} }};

// { a: { b: 1, c: 1, d: { e: 1, f: 1 } } }
// f should be 'f'
```

lib/deep_merge.js
```diff
@@ -0,0 +1,17 @@
...
+
+  // Iterate through `source` properties and if an `Object` set property to merge of `target` and `source` properties
+  for (const key of Object.keys(source)) {
+    if (source[key] instanceof Object) Object.assign(source[key], deepMerge(target[key], source[key]));
+  }
+
+  // Join `target` and modified `source`
+  Object.assign(target || {}, source);
```
curbengh: I don't think `|| {}` is necessary since `target` is never a null; even if `target` is null, `|| {}` doesn't prevent error.

SukkaW: It is used to handle if `target` is `undefined`, so `deepMerge(obj)` will return a shallow copy of `obj`.

lib/deep_merge.js
SukkaW: https://github.com/hexojs/hexo-util/blob/d652ff12092c96b34908a90a2207731bbba30865/test/deep_merge.spec.js#L34

Since the unit test passed, I believe there is no problem here.

curbengh: The unit test didn't run, I suggest not to rely on it entirely. I tested locally and can verify that currently `f` key has value of `1` not `'f'`.

lib/deep_merge.js
curbengh: When `deepMerge(obj)`, `obj` is passed as `target` parameter, so when `obj` is `undefined`, `deepMerge(obj)` it will return `undefined`.

Anyhow, since L4 needs to be removed, `|| {}` is indeed necessary to avoid error when `target` is `undefined`.

curbengh:
``` js
function deepMerge(target, source) {
  target = Object.assign({}, target);
  source = Object.assign({}, source);

  // Iterate through `source` properties and if an `Object` set property to merge of `target` and `source` properties
  for (const key of Object.keys(source)) {
    if (source[key] instanceof Object) Object.assign(source[key], deepMerge(target[key], source[key]));
  }

  // Join `target` and modified `source`
  Object.assign(target || {}, source);
  return target;
}

const obj1 = {a: {b: 1, c: 1, d: {e: 1, f: 1}}};
const obj2 = {a: {b: 2, d: {f: 'f'} }};

deepMerge(obj1,obj2)

console.log(deepMerge(obj1,obj2))
// { a: { b: 2, d: { f: 'f', e: 1 }, c: 1 } }

console.log(deepMerge(obj1))
// { a: { b: 1, c: 1, d: { e: 1, f: 1 } } }
```

Is it intended that `obj1` is not modified? In that case, the example in the docs needs to be updated to,

``` js
const obj1 = {a: {b: 1, c: 1, d: {e: 1, f: 1}}};
const obj2 = {a: {b: 2, d: {f: 'f'} }};
const obj3 = deepMerge(obj1, obj2);

console.log(obj3);
// {a: {b: 2, c: 1, d: {e: 1, f: 'f'} }}
```

curbengh:
The [deepmerge](https://github.com/TehShrike/deepmerge/blob/master/index.js) package works well, it doesn't modify source and target objects. We can derive from it with relevant es6-ification.

SukkaW:
@curbengh 

https://runkit.com/sukkaw/5ddaa73ff19d1a001c7b51e6

The benchmark shows the `deepmerge` package is faster..

SukkaW:
@curbengh I believe the default option of `deepmerge` will be good enough. I might update the test (https://github.com/hexojs/hexo/pull/3880) to see if the default options works.

curbengh:
Using the example in the readme, somehow source is also modified.

``` js
const obj1 = {a: {b: 1, c: 1, d: {e: 1, f: 1}}};
const obj2 = {a: {b: 2, d: {f: 'f'} }};
deepMerge(obj1, obj2);

console.log(obj1) // { a: { b: 2, d: { f: 'f', e: 1 }, c: 1 } }
console.log(obj2) // { a: { b: 2, d: { f: 'f', e: 1 }, c: 1 } }
```

But the example used in the unit test is fine,

``` js
const obj1 = {a: 0, b: 1, c: {d: 1}, e: 4};
const obj2 = {b: 3, c: {d: 2}};
deepMerge(obj1, obj2);

console.log(obj1) // { a: 0, b: 3, c: { d: 2 }, e: 4 }
console.log(obj2) // { b: 3, c: { d: 2 } }
```

SukkaW:
@curbengh I know what happend. We all know that `Object.assign({}, some obj)` only does a shallow copy of `obj`. It means that nested properties of `src` will still be **copied by reference**. Thus in the case `source` obj has been touched.

Without a deep clone, we can not achieve leave `source` completely untouched with `Object.assign`.

SukkaW:
https://repl.it/repls/ThoroughIndigoAddons

According to the benchmark, `deepmerge` is only a little bit slower than `lodash.merge`. I believe it is because `arrayMerge`.

README.md
```diff
@@ -111,6 +112,28 @@ decodeURI(format(new URL('http://xn--br-mia.com.com/b%C3%A1r'), {unicode: true})
 // http://bár.com/báz
 ```
 
+### deepMerge(target, source)
+
+Merges the enumerable properties of two objects deeply while neither `target` or `source` is modified.
```
curbengh:
```
Merges the enumerable properties of two objects deeply. `target` and `source` remain the same.
```

README.md
```diff
@@ -111,6 +112,28 @@ decodeURI(format(new URL('http://xn--br-mia.com.com/b%C3%A1r'), {unicode: true})
...
+// Merge deeply
+const obj1 = {a: {b: 1, c: 1, d: {e: 1, f: 1}}};
+const obj2 = {a: {b: 2, d: {f: 'f'} }};
+
+deepMerge(obj1, obj2);
+// {a: {b: 2, c: 1, d: {e: 1, f: 'f'} }}
+```
+
+``` js
+// Objects will be combined at the same index in the two arrays, just as lodash.merge
```
curbengh: Arrays will be combined in the same property, similar to lodash.merge

curbengh:
I'm a little surprised `chai-should.eql()` also works on object, I previously assumed it's similar to `===` which [doesn't work](https://stackoverflow.com/q/1068834) on object.

PR #146, opened 2019-12-01
refactor: use alias package punycode.js

SukkaW:
`punycode` package uses the same name as the inbuilt punycode package in node.js. Related report can be found [here](https://github.com/bestiejs/punycode.js/issues/97).

And I have found the package [`punycode.js`](https://www.npmjs.com/package/punycode.js) on npm. It is exact the same package as `punycode` and has the same maintainer, which could act as an alias to `punycode`. The only difference between them is `punycode` package has added es6 module support in `punycode@2.1.1`.


curbengh:
whatwg has built-in punycode function, but currently it's not used here yet due to a bug in Node 8. Once Node 8 EOL, punycode shall be dropped.

https://github.com/hexojs/hexo-util/blob/ea43fba3635c50f566ea18ceff2fe7d3e87f90d4/lib/encode_url.js#L23-L24

PR #147, opened 2019-12-02
fix(url_for/full_url_for): do not process absolute url

curbengh:
Noticed an edge case while testing https://github.com/hexojs/hexo-renderer-marked/pull/124

``` js
const ctx = {
  config: {
    url: 'http://example.com',
    root: '/goal/'
  }
}

console.log(url_for.call(ctx, 'http://example.com/hash/bob.html'))
// /goal/http:/example.com/hash/bob.html
```

---

`ReExp.test()` is slightly faster than `String.startsWith()/String.endsWith()` and easier to add a new pattern.

### DIFF
--- a/lib/full_url_for.js
+++ b/lib/full_url_for.js
@@ -4,7 +4,8 @@ const { parse, URL } = require('url');
 const encodeURL = require('./encode_url');
 
 function fullUrlForHelper(path = '/') {
-  if (path.startsWith('//')) return path;
+  const pathRegex = /^(\/\/|http(s)?:)/;
+  if (pathRegex.test(path)) return path;
 
   const { config } = this;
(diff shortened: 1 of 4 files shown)


lib/full_url_for.js
```diff
@@ -4,7 +4,8 @@ const { parse, URL } = require('url');
 const encodeURL = require('./encode_url');
 
 function fullUrlForHelper(path = '/') {
-  if (path.startsWith('//')) return path;
+  const pathRegex = /^(\/\/|http(s)?:)/;
```
curbengh: not sure if `ignore case` flag is needed, I don't mind adding it. I benchmarked the flag and I didn't notice any difference.

PR #152, opened 2019-12-05
feat(pretty_urls): bring up

SukkaW:
Bring up a `prettyUrls()` to rewrite urls to pretty URLs. Used to remove duplicated code introduced in https://github.com/hexojs/hexo/pull/3691 & https://github.com/hexojs/hexo/pull/3893.

Since `pretty_urls` is already applied for `url_for` and `full_url_for` in #143, this PR utilize `prettyUrls()` for `url_for` and `full_url_for` as well.

Also, https://github.com/hexojs/hexo/pull/3893 should be closed & https://github.com/hexojs/hexo/pull/3691 should be revert as `full_url_for` already has built-in `pretty_urls` support. I might open another PR to update the related code & test after newer version of `hexo-util` is released.


README.md
```diff
@@ -360,6 +361,39 @@ permalink.stringify({year: '2014', month: '01', day: '31', title: 'test'})
...
 ```
 
+### prettyUrls(url, [options])
+
+Rewrite urls to pretty URLs.
+
+Option | Description | Default
+--- | --- | ---
+`trailing_index` | `/about/index.html -> /about/` when `false` | `true`
+`trailing_html` | `/about.html -> /about` when `false` | `true`
```
curbengh: Since `trailing_index` retain trailing slash by default, I think `trailing_html` should retain it as well. If a trailing slash is not desired in _both_ `trailing_index` and `trailing_html`, can disable `trailing_slash` (suggest to introduce this option in this PR too).

README.md
```diff
@@ -360,6 +361,39 @@ permalink.stringify({year: '2014', month: '01', day: '31', title: 'test'})
...
+
+Notice if only `trailing_html` is enabled, trailing `index.html` will remains the same (not to be rewritten to `index`).
```
curbengh: Can be simplified to

```
Note: `trailing_html` ignores any link with a trailing `index.html`.
```

README.md
SukkaW: For most web server (like nginx, caddy), when they receive a URL like `/foo/bar/`, they will looks up html files (`index.html`, `index.htm`, `index.php`, `index.asp`, etc.) under `foo/bar/` directory. But when receive a URL like `foo/bar`, they will look up `bar.html`, `bar.htm` under `/foo/` directory.

As you can see, both `https://hexo.io/docs/permalinks` and `https://hexo.io/docs/permalinks.html` will return `200 OK` while `https://hexo.io/docs/permalinks/` will be 301 redirect to `https://hexo.io/docs/permalinks`.

README.md
SukkaW: Updated.

README.md
curbengh: you're right, I mistook netlify's pretty url. Upon checking,

`Rewrite link URLs to pretty URLs (/about.html -> /about, /about/index.html -> /about/)`

Issue #145, opened 2019-11-30
The new code block layout made it difficult to select and copy code without including the line number


ppoffice:
Before the new layout:
I can select **code and code only** by hand or using `clipboard.js`
![image](https://user-images.githubusercontent.com/8849362/69896221-b7a70b00-130a-11ea-96fd-19ba63569dac.png)

After the new layout:
Line number is also selected as each number is within the same line as the code
![image](https://user-images.githubusercontent.com/8849362/69896289-a1e61580-130b-11ea-8b94-2184cd87363b.png)

Related code:
https://github.com/hexojs/hexo-util/blob/master/lib/highlight.js#L45-L55

SukkaW:
Related issue: https://github.com/hexojs/hexo/issues/3884 https://github.com/hexojs/hexo/issues/3913 https://github.com/hexojs/hexo-util/commit/401dde1621331d07ce52caf37c94951bac33ddb7#r36250206

@curbengh @stevenjoezhang @jiangtj

Since it is a serious Breaking Changes that already breaks a lot of theme. I suggest to revert the PR https://github.com/hexojs/hexo-util/pull/132 and release an emergency release of hexo-util & hexo.

SukkaW:
SInce the `prismjs` is already on the roadmap, IMHO we shouldn't bring any changes to current `highlightjs` since it will be deprecated soon.

curbengh:
https://github.com/hexojs/hexo-util/pull/153

PR #132, opened 2019-11-15
feat(highlight): adjust the layout of the highlight code block

stevenjoezhang:
## Current behavior
```html
<tr>
  <td class="gutter"><pre>
    <span class="line">1</span>
    <br>
    <span class="line">2</span>
    <br>
    <span class="line">3</span>
  </pre></td>

  <td class="code"><pre>
    <span class="line">hello, world</pre>
    <br>
    <span class="line">hello, world</span>
    <br>
    <span class="line">hello, world</span>
  </pre></td>
</tr>
```

## New behavior
```html
<tr>
  <td class="gutter"><pre>
    <span class="line">1</span>
  </pre></td>
  <td class="code"><pre>
    <span class="line">hello, world</pre>
  </td>
</tr>

<tr>
  <td class="gutter"><pre>
    <span class="line">2</span>
  </pre></td>
  <td class="code"><pre>
    <span class="line">hello, world</pre>
  </td>
</tr>

<tr>
  <td class="gutter"><pre>
    <span class="line">3</span>
  </pre></td>
  <td class="code"><pre>
    <span class="line">hello, world</pre>
  </td>
</tr>
```

PR #153, opened 2019-12-09
revert: feat(highlight): adjust the layout of the highlight code block

SukkaW:
In this PR I just simply revert #132 because #132 introduces just too much Breaking Changes, see discussion at https://github.com/hexojs/hexo-util/pull/132 and related issue #145. 

IMHO, we should not bring any style changes in highlightjs any more. It will break too much things, especially when we are going to deprecate highlightjs and bring up prismjs.

Hexo has nearly 300 themes, "NexT theme works fine" is just not an enough excuse.

cc @jiangtj @stevenjoezhang @ppoffice

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -2,7 +2,6 @@
 
 const hljs = require('highlight.js');
 const alias = require('../highlight_alias.json');
-
 const escapeHTML = require('./escape_html');
 
 function highlightUtil(str, options = {}) {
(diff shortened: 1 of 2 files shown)


SukkaW:
Also, we need to merge this PR and a release a newer version as soon as possible. Hexo needs a emergency release to include this PR and https://github.com/hexojs/hexo-util/pull/147

PR #158, opened 2019-12-16
feat(cache): enable lru cache

SukkaW:
I have setup a stat by adding a few line of the code to `node_modules/hexo` in my local dummy site:

```js
let num = 0;

// .....
num++;
console.log(num);
```

It turns out that even as few as 17 posts in [hexo-theme-unit-test](https://github.com/hexojs/hexo-theme-unit-test), `full_url_for()` helper will be called for 537 times and `isExternalLink()` will be called for 2599 times in one generation!

So I bring up this PR. The `node-lfu-cache` is enabled for those functions:

- full_url_for()
- gravatar()
- is_external_link()
- relative_url()
- url_for()

PR #162, opened 2019-12-18
feat(cache): plain object cache

SukkaW:
Context: #158 

Cache with plain object.

cc @dailyrandomphoto @seaoak

### DIFF
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -1,0 +1,32 @@
+'use strict';
+
+module.exports = class Cache {
+  constructor() {
+    this.cache = {};
+  }
+
+  set(id, value) {
+    this.cache[id] = value;
+  }
+
+  has(id) {
+    return this.cache[id] != null;
+  }
+
+  get(id) {
+    return this.cache[id];
+  }
+
+  apply(id, value) {
+    if (this.has(id)) return this.get(id);
+
+    if (typeof value === 'function') value = value();
+
+    this.set(id, value);
+    return value;
+  }
+
+  flush() {
+    this.cache = {};
+  }
+};
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -10,7 +10,7 @@ module.exports = class Cache {
   }
 
   has(id) {
-    return this.cache[id] != null;
+    return typeof this.cache[id] !== 'undefined';
   }
 
   get(id) {
--- a/lib/cache.js
+++ b/lib/cache.js
@@ -17,6 +17,10 @@ module.exports = class Cache {
     return this.cache[id];
   }
 
+  del(id) {
+    delete this.cache[id];
+  }
+
   apply(id, value) {
     if (this.has(id)) return this.get(id);
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Utilities for [Hexo].
 
 - [Installation](#installation)
 - [Usage](#usage)
+- [Cache](#cache)
 - [CacheStream](#cachestream)
 - [camelCaseKeys](#camelcasekeysobj-options)
 - [createSha1Hash](#createsha1hash)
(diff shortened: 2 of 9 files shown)


SukkaW:
Benchmark result:

| Node.js Version | | Cold Process Time | Memory |
|:---:|:---:|:---:|:---:|
| 8 | Current master branch of Hexo | 28.3s | 587MB |
| 8 | Hexo with sukkaw/hexo-util#plain-object-cache | 26.3s | 589MB |
| 10 |  Current master branch of Hexo | 19.3s | 596MB |
| 10 | Hexo with sukkaw/hexo-util#plain-object-cache | 21.7s | 600MB |
| 12 |  Current master branch of Hexo | 19.0s | 582MB |
| 12 | Hexo with sukkaw/hexo-util#plain-object-cache | 16.8s | 580MB |
| 13 |  Current master branch of Hexo | 24.9s | 587MB |
| 13 | Hexo with sukkaw/hexo-util#plain-object-cache | 22.1s | 590MB |

curbengh:
I notice repetitive codes, perhaps can refactor into a `cache.js`?

SukkaW:
@curbengh I see no necessary.

I have bring up something like this:

```js
// cache.js

'use strict';

module.exports = class Cache {
  constructor() {
    this.cache = {};
  }

  set(id, value) {
    this.cache[id] = value;
  }

  has(id) {
    return this.cache[id] != null;
  }

  get(id) {
    return this.cache[id];
  }

  flush() {
    this.cache = {};
  }
};
```

But the usage will end in something like:

```js
const Cache = require(./cache);
const cache = new Cache();

if (cache.has(cacheId)) return cache.get(cacheId);
...
cache.set(cacheId, value);
```

Nothing changed...

curbengh:
``` js
const Cache = require('./cache');
const cache = new Cache();

if (cache.has(cacheId)) return cache.get(cacheId);
...
cache.set(cacheId, value);
```

while it doesn't reduce the line, it does look more readable.

SukkaW:
@curbengh I have bring up `cache.js` with `set()`, `get()`, `has()`, `flush()` and `apply()` methods.

SukkaW:
Performance with `Cache()`:

| Node.js Version | | Cold Process Time | Memory |
|:---:|:---:|:---:|:---:|
| 8 | Current master branch of Hexo | 25.2s | 591MB |
| 8 | Hexo with sukkaw/hexo-util#plain-object-cache | 20.0s | 603MB |
| 10 |  Current master branch of Hexo | 18.7s | 593MB |
| 10 | Hexo with sukkaw/hexo-util#plain-object-cache | 22.9s | 585MB |
| 12 |  Current master branch of Hexo | 18.4s | 576MB |
| 12 | Hexo with sukkaw/hexo-util#plain-object-cache | 17.2s | 584MB |
| 13 |  Current master branch of Hexo | 18.8s | 642MB |
| 13 | Hexo with sukkaw/hexo-util#plain-object-cache | 16.9s | 624MB |

lib/cache.js
```diff
@@ -0,0 +1,36 @@
...
+  constructor() {
+    this.cache = {};
+  }
+
+  set(id, value) {
+    this.cache[id] = value;
+  }
+
+  has(id) {
+    return this.cache[id] != null;
```
seaoak: This line will intend:

```js
    return this.cache[id] !== undefined;
```

And also, maybe be better:

```js
    return this.cache.hasOwnProperty(cacheId);
```

The later works well with `delete` operator, too.

SukkaW: https://github.com/hexojs/hexo/blob/fa59eb28e8344c84f7fa17d4ee48f8def79ce07e/lib/plugins/helper/fragment_cache.js#L10

`!= null` is used in fragment_cache helper and it will return `false` when meets `undefined`.

SukkaW: https://repl.it/repls/PertinentEffectiveLicenses

I have setup a repl and find out `hasOwnProperty` is just too slow.

curbengh: > Unless we want to cache null value, otherwise I believe != null will work well.

I think there might be edge case where `null` value is intended to be cached. Checking for undefined is more accurate.

SukkaW: @curbengh @seaoak 

I have updated to use `id in this.cache` which makes `undefined` & `null` both cacheable.

seaoak:
I tasted this patch in a large blog (dummy 3000 posts).

The result shows this patch is 13% faster than `master`.
(no regression!)

---
## Details:

Set CPU affinity `0x00010000` and priority `HIGH` for `hexo` command (using `START` command).
For each condition, measure 10 times and calculate the average.
Before each measurement, do `hexo clean` and `rm debug.log`.

### Result

Elapsed time:

Branch | Average | Spread (worst - best)
-------|---------|-----------------------
`master` | 77.34sec | 1.67sec
`SukkaW` | 67.15sec (-13.18%) | 2.44sec

### Source code:

- `master` => hexojs/hexo-util@621c9e3d73e89790faa24a1b0c641dda1074752f
- `SukkaW` => SukkaW/hexo-util@afdc7f89602249941252f69b5d41fa82930f6734

### Environment:

- OS: Windows 10 64bit
- CPU: Ryzen Threadripper 2950X (16core/32thread, 3.5GHz)
- MEM: 32GB (8GB x4) DDR4 (3600MHz)
- SSD: Samsung EVO970 (1TB)
- Node v12.14.0

seaoak:
> Measure elapsed time of `hexo generate --debug`.

Memory usage is about 2.3GB as the worst case in both conditions.
(measured with `tasklist` command)

PR #166, opened 2019-12-24
fix(tocObj): avoid checking length of null

curbengh:
tocObj currently crashes hexo if the heading doesn't have `id` attribute, e.g. `<h1>Title 1</h1>`. This PR makes it more robust.

With this PR, in relation to the [`toc()`](https://github.com/hexojs/hexo/blob/f50821c596a9c3a53619cbb846509e6071a1780d/lib/plugins/helper/toc.js#L50) helper, it would output

```
<a class="header-link" href="#null">
```

obviously anchor link wouldn't work, but at least there is a toc.

### DIFF
--- a/lib/toc_obj.js
+++ b/lib/toc_obj.js
@@ -11,7 +11,7 @@ const parseHtml = html => {
 const getId = ele => {
   const { id } = ele.attribs;
   const { parent } = ele;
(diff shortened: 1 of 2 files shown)


curbengh:
I intend to bump patch after merging this.

PR #168, opened 2019-12-25
feat: bring up prism highlight support

### LINKED ISSUE #108: Feature request: support other syntax highlight libraries

SukkaW:
Related to https://github.com/hexojs/hexo/issues/1036  https://github.com/hexojs/hexo/issues/4010

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,7 +445,7 @@ Syntax highlighting for a code block using PrismJS.
 Option | Description | Default
 --- | --- | ---
 `lineNumber` | Whether to show line numbers | true
-`language` | Language | `'plain'`
+`lang` | Language | `'none'`
 `tab`| Replace tabs |
 
 ### relative_url(from, to)
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ Utilities for [Hexo].
 - [Pattern](#patternrule)
 - [Permalink](#permalinkrule-options)
 - [prettyUrls](#prettyurlsurl-options)
+- [prismHighlight](#prismhighlightstr-options)
 - [relative_url](#relative_urlfrom-to)
 - [slugize](#slugizestr-options)
 - [spawn](#spawncommand-args-options)
@@ -436,6 +437,16 @@ prettyUrls('/foo/bar.html', { trailing_index: false, trailing_html: false });
 prettyUrls('/foo/bar/index.html', { trailing_index: false, trailing_html: false });
 // /foo/bar/
 ```
+
+### prismHighlight(str, [options])
+
+Syntax highlighting for a code block using PrismJS.
+
+Option | Description | Default
+--- | --- | ---
+`lineNumber` | Whether to show line numbers | true
+`language` | Language | `'plain'`
+`tab`| Replace tabs |
 
 ### relative_url(from, to)
 
--- a/README.md
+++ b/README.md
@@ -447,6 +447,9 @@ Option | Description | Default
 `lineNumber` | Whether to show line numbers | true
 `lang` | Language | `'none'`
 `tab`| Replace tabs |
+`isPreprocess` | Enable preprocess or not | true
+
+When `isPreprocess` is enabled, `prismHighlight()` will return PrismJS processed HTML snippet. Otherwise `str` will only be escaped and `prismHighlight()` will return the HTML snippet that is suitable for `prism.js` working in the Browser.
 
 ### relative_url(from, to)
 
--- a/lib/prism.js
+++ b/lib/prism.js
@@ -41,7 +41,7 @@ function PrismUtil(str, options = {}) {
 
   const {
     lineNumber = true,
-    language
+    language = 'plain'
   } = options;
 
   const preTagClassArr = [];
@@ -52,13 +52,18 @@ function PrismUtil(str, options = {}) {
   const startTag = `<pre class="${preTagClassArr.join(' ')}"><code class="language-${language}">`;
   const endTag = '</code></pre>';
 
-  const highlightedCode = prismHighlight(str, language);
-
-  if (lineNumber) {
-    return startTag + highlightedCode + lineNumberUtil(highlightedCode) + endTag;
+  let parsedCode = '';
+  if (language === 'plain') {
+    parsedCode = prismHighlight(str, language);
+  } else {
+    parsedCode = escapeHTML(str);
   }
 
-  return startTag + highlightedCode + endTag;
+  if (lineNumber) {
+    parsedCode += lineNumberUtil(parsedCode);
+  }
+
+  return startTag + parsedCode + endTag;
 }
 
 module.exports = PrismUtil;
--- a/lib/prism.js
+++ b/lib/prism.js
@@ -1,0 +1,64 @@
+'use strict';
+
+const Prism = require('prismjs');
+const prismLoadLanguages = require('prismjs/components/');
+
+const escapeHTML = require('./escape_html');
+const unescapeHTML = require('./unescape_html');
+
+/**
+ * Wrapper of Prism.highlight()
+ * @param {String} code
+ * @param {String} language
+ */
+function prismHighlight(code, language) {
+  // Prism has not load the language pattern
+  if (!Prism.languages[language]) prismLoadLanguages(language);
+
+  if (Prism.languages[language]) {
+    // Prism escapes output by default
+    return Prism.highlight(unescapeHTML(code), Prism.languages[language], language);
+  }
+
+  // Current language is not supported by Prism, return origin code;
+  return escapeHTML(code);
+}
+
+/**
+ * Generate line number required HTML snippet
+ * @param {String} code - Highlighted code
+ */
+function lineNumberUtil(code) {
+  const matched = code.match(/\n(?!$)/g);
+  const num = matched ? matched.length + 1 : 1;
+  const lines = new Array(num + 1);
+
+  return `<span aria-hidden="true" class="line-numbers-rows">${lines.join('<span></span>')}</span>`;
+}
+
+function PrismUtil(str, options = {}) {
+  if (typeof str !== 'string') throw new TypeError('str must be a string!');
+
+  const {
+    lineNumber = true,
+    language
+  } = options;
+
+  const preTagClassArr = [];
+
+  if (lineNumber) preTagClassArr.push('line-numbers');
+  preTagClassArr.push(`language-${language}`);
+
+  const startTag = `<pre class="${preTagClassArr.join(' ')}"><code class="language-${language}">`;
+  const endTag = '</code></pre>';
+
+  const highlightedCode = prismHighlight(str, language);
+
+  if (lineNumber) {
+    return startTag + highlightedCode + lineNumberUtil(highlightedCode) + endTag;
+  }
+
+  return startTag + highlightedCode + endTag;
+}
+
+module.exports = PrismUtil;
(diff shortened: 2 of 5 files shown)


SukkaW:
Unlikes `highlight.js`, PrismJS has only a little support for Node.js. Features supported by `highlight.js` (like `mark` and `firstLine`) will only available for PrismJS works in a Browser.

SukkaW:
Coverage for `prism.js` is 100%, yeah!

lib/prism.js
```diff
@@ -0,0 +1,89 @@
...
+const escapeHTML = require('./escape_html');
+const unescapeHTML = require('./unescape_html');
+
+/**
+ * Wrapper of Prism.highlight()
+ * @param {String} code
+ * @param {String} language
+ */
+function prismHighlight(code, language) {
+  console.log(Object.keys(Prism.languages));
```
curbengh: left from debugging?

lib/prism.js
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Generate line number required HTML snippet
+ * @param {String} code - Highlighted code
+ */
+function lineNumberUtil(code) {
+  const matched = code.match(/\n(?!$)/g);
+  const num = matched ? matched.length + 1 : 1;
+  const lines = new Array(num + 1);
+
+  return `<span aria-hidden="true" class="line-numbers-rows">${lines.join('<span></span>')}</span>`;
```
curbengh: This approach (instead of table) is used to support prismjs css layout? Or it's recommended by prismjs?
Can this support https://github.com/hexojs/hexo-util/issues/145?

README.md
```diff
@@ -437,6 +438,16 @@ prettyUrls('/foo/bar/index.html', { trailing_index: false, trailing_html: false
...
 ```
 
+### prismHighlight(str, [options])
+
+Syntax highlighting for a code block using PrismJS.
+
+Option | Description | Default
+--- | --- | ---
+`lineNumber` | Whether to show line numbers | true
+`language` | Language | `'plain'`
```
curbengh: Prefer to use shorter `lang` like in `highlight()`.

lib/prism.js
SukkaW: > This approach (instead of table) is used to support prismjs css layout?

Specifically, `prism-line-numbers.css`: https://github.com/PrismJS/prism/blob/master/plugins/line-numbers/prism-line-numbers.css

PrismJS is using some CSS3 feature to render line number.

lib/prism.js
SukkaW: Emmmm. I am still testing if PrismJS will load language files repeatedly.

lib/prism.js
```diff
@@ -0,0 +1,89 @@
...
+
+function PrismUtil(str, options = {}) {
+  if (typeof str !== 'string') throw new TypeError('str must be a string!');
+
+  const {
+    lineNumber = true,
+    language = 'plain',
+    tab
+  } = options;
+
```
seaoak: Is is necessary to translate the language name `plain` to `none` ?

In Hexo (with `highlightjs`), the language name `plain` means "no highlighting".
By the document of Prism, it is achieved by using the class name `language-none`.
https://prismjs.com/#basic-usage

I think the manner in Hexo should keep regardless what highlighter is used.
So above translation will be desirable.

lib/prism.js
```diff
@@ -0,0 +1,89 @@
...
+ * @param {String} language
+ */
+function prismHighlight(code, language) {
+  console.log(Object.keys(Prism.languages));
+  // Prism has not load the language pattern
+  if (!Prism.languages[language] && prismSupportedLanguages.includes(language)) prismLoadLanguages(language);
+
+  if (Prism.languages[language]) {
+    // Prism escapes output by default
+    return Prism.highlight(unescapeHTML(code), Prism.languages[language], language);
```
seaoak: If `escapeHTML(code)` in the later line is necessary, `code` will be a plain text (no escaped).
So I thnk this `unescapeHTML()` will not be necessary.

lib/prism.js
```diff
@@ -0,0 +1,89 @@
...
+
+  // Current language is not supported by Prism, return origin code;
+  return escapeHTML(code);
+}
+
+/**
+ * Generate line number required HTML snippet
+ * @param {String} code - Highlighted code
+ */
+function lineNumberUtil(code) {
```
seaoak: @SukkaW Do you intend to re-implement the "line-numbers" plugin of Prism?
If yes, some functions seem to lack (such as a SPAN element with the class name "line-numbers-sizer").

Could you taste a work-around to use plugins in Node.js ?
https://github.com/PrismJS/prism/issues/1171#issuecomment-470929808

SukkaW: https://github.com/PrismJS/prism/issues/1171#issuecomment-381774025

> Line Numbers: Does not work and returns early. Requires DOM.

And the current approach should be compatible with `prism-line-numbers.css`:

https://github.com/PrismJS/prism/blob/master/plugins/line-numbers/prism-line-numbers.css

lib/prism.js
SukkaW: We cannot identified whether the `code` is escaped or not, so unescape the code before highlighting, and always return the escaped result.

lib/prism.js
SukkaW: I will update to that.

lib/prism.js
SukkaW: https://prismjs.com/plugins/line-numbers/

image

As you can see, this is the DOM structure after enabling the prism's line numbers plugin. No `.line-numbers-sizer`.

lib/prism.js
SukkaW: I have updated the default `lang` to `none` to be consistent with `prismjs`.

`plain` is still acceptable, while `{ lang: 'plain' }` will result in `.language-none`.

seaoak: Thanks. I recognized!

lib/prism.js
seaoak: `unescapeHTML()` breaks a sample HTML code such as `"&" => &amp;`.
(The result will look like `"&" => &`.  This will not be intended by the author.)
I think we should believe that `code` is a plain text.

lib/prism.js
```diff
@@ -0,0 +1,92 @@
...
+  const matched = code.match(/\n(?!$)/g);
+  const num = matched ? matched.length + 1 : 1;
+  const lines = new Array(num + 1);
```
seaoak: I think this code will cause misreading.
The variable `lines` should be a sequence of `<span></span>`.
The code `new Array(num + 1)` is only a trick to use an edge case of `Array.prototype.join()`.
Please keep original codes as you can.
https://github.com/PrismJS/prism/blob/master/plugins/line-numbers/prism-line-numbers.js

SukkaW: Updated and rebased.

lib/prism.js
SukkaW: `unescapeHTML` is removed and rebased.

lib/prism.js
SukkaW: image

https://github.com/PrismJS/prism/blob/e20a2b711a0cd4ee7e6b49b07f7ce165df2a7980/plugins/line-numbers/prism-line-numbers.js#L48-L49

It seems that the `lineNumberSizer` will be removed.

lib/prism.js
SukkaW: In `prism-line-numbers.css` line numbers are applied with `user-select: none`.

After then, it depends on the theme maintainer.

lib/prism.js
```diff
@@ -0,0 +1,92 @@
...
+  preTagClassArr.push(`language-${language}`);
+
+  if (tab) str = replaceTabs(str, tab);
+
+  const startTag = `<pre class="${preTagClassArr.join(' ')}"><code class="language-${language}">`;
+  const endTag = '</code></pre>';
+
+  let parsedCode = '';
+
+  if (language === 'none' || !isPreprocess) {
```
SukkaW: https://github.com/hexojs/hexo/issues/4010#issuecomment-568906344

> During [hexojs/hexo-util#168](https://github.com/hexojs/hexo-util/pull/168), I have noticed PrismJS is mainly designed for working at browser side. Many features likes "highlight specific lines", "The first line number" only work in browser.
> 
> So if we are going to bring PrismJS to Hexo, IMHO we should address two options:
> * `browser` mode. In this mode Hexo will produce HTML snippet that will could be processed by `prism.js` & `prism.css` in browser. The full features of PrismJS could be used then, and it will be compatible with all options of current `include_code()` and `code()` helper.
> * `preprocess` mode. In this mode Hexo will produce prism-parsed HTML snippet. Then theme/user only need to add `prism.css`. However, many features will no longer works under `preprocess` mode thus will not be compatible with `include_code()` and `code()` helper.

lib/prism.js
seaoak: I agreed.

seaoak:
Just in case...

I think whether the result should have a *SIGN* which shows the code snippet is for Prism.

The result of existing `highlight()` includes an HTML class name `highlight` or `hljs`.
This shows the code snippet is for `highlightjs` clearly.

How `prism.js` on browser does find the target `<pre>` elements ?

SukkaW:
@seaoak 

https://prismjs.com/#basic-usage

Here is the basic guide of PrismJS. It seems that we only need to add `language-xxxx` or `language-xxxx` to `<code>` and `prism.js` will do the rest. `prism.js` will also add `language-xxxx` to parent `<pre>`.

An example on codepen: https://codepen.io/eksch/pen/jukqf

seaoak:
@SukkaW I understand. LGTM.

Issue #164, opened 2019-12-21
`highlight()` is inconsistent when code is illegal syntax


seaoak:
Generally the result of `highlight()` contains lauguage name as HTML class name.

If code is illegal syntax for the specified language,
the HTML class name is added or not (inconsistent).

option `lang` | illegal syntax? | option `hljs` | class name of outer `<figure>` | class name of inner `<code>`
----------------|---------------|---------------|--------------------------------|----------------------
`java` | no  | `false` | `highlight java` | N/A (none)
`java` | no  | `true`  | `highlight java` | `hljs java`
`java` | yes | `false` | `highlight` | N/A (none)
`java` | yes | `true`  | `highlight` | `hljs java` 
none   | - | `false` | `highlight plain` | N/A (none)
none   | - | `true`  | `highlight plain` | `hljs plain` 

I think we have three options:

 1. always add laugage name even if illegal syntax.
 2. never add language name if illegal syntax.
 3. add the language name `plain` if illegal syntax.

I think Option 1 might be best because users will intend to be.

Please advice!

---
I think this inconsistency is caused by undocumented behavior of `highlight()` of `highlightjs` package.

The API document of `highlightjs` package is:
https://highlightjs.readthedocs.io/en/latest/api.html#highlight-languagename-code-ignore-illegals-continuation

According to above API document, `highlight()` throws an exception in case of detecting illegal syntax for the language.
But actually not throw any exception (regardless `ignore_illegals` parameter).
Instead, `illegal` property of the return value is set to `true` and
`language` property of the return value does not exist.

Current code of `lib/highlight.js` considers the exception halfway.
And the lack of `lanaguage` property is not expected.

Anyway the code should be modified to keep consistency.

p.s. This issue is derived from hexojs/hexo#3984

stevenjoezhang:
@seaoak I agree that option 1 might be the best choice. Can you help test whether this pull request meets your expectations? https://github.com/hexojs/hexo-util/pull/246

PR #30, opened 2017-12-11
hljs - Added compatibility for CSS styles.

alxbl:
This pull request adds additional support to render the highlighting with `hljs-` prefix and preserves backward compatibility for existing themes/stylesheets.

Support for both marked lines and line numbers, as well as vanilla highlight.js blocks is included.

This is a quick proof of concept that I whipped up after I started working on my theme and realizing that hljs CSS themes weren't supported. I hope it can be helpful to hexo users.

It's a two part pull request as [hexo/#2901](https://github.com/hexojs/hexo/pull/2901) adds support for propagating the `hljs` configuration through to the highlight plugin.

If everything is good (or after necessary changes), I will be updating the documentation as well to document the new config for `_config.yml`.

Cheers,

PR #171, opened 2019-12-29
Fix highlighting when hljs=true

ppwwyyxx:
I found that the highlighting is incorrect when options.hljs=true.

It should use the `tryHighlight` function instead, which handles multi-line codeblock.

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -116,8 +116,6 @@ function highlight(str, options) {
     return result;
   }
 
-  if (options.hljs) return hljs.highlight(lang, str);
-
   return tryHighlight(str, result.language) || result;
 }
 


seaoak:
> I found that the highlighting is incorrect when options.hljs=true.

Could you show an example such case ?

Certainly, the existing code seems something strange.
But I think it works enough (already used by many users).
Of cource, it will works well for a multi-line codeblock.

I don't know the reason why this `if()` statement was introduced.
So I cannot decide to remove this `if()` statement.

Can you explain ?

---
BTW,

The API document of `highlightjs` package is:
https://highlightjs.readthedocs.io/en/latest/api.html#highlight-languagename-code-ignore-illegals-continuation

According to above API document, `highlight()` throws an exception in case of detecting illegal syntax for the language.
But existing code does not consider it.

I think `try {...} catch() {...}` statement will be necessary here
(in a similar way to `tryHighlight()`).

Please refer to issue #164.

ppwwyyxx:
Thanks for the response. I thought it cannot highlight multi-line block, but it turns out that, it simply cannot highlight certain multi-line block:
```
> str =  'class B {\n' +
...     '  public:\n' +
...     '    explicit B(const A&amp; a) : a_{a} {}\n' +   
...     '    const A&amp; a_;\n' +                        
...     '};\n' +                                                                                                   
...     '\n' +                                           
...     'int main() {\n' +                               
...     '  A a; B b{a};\n' +
...     '  printf("HERE\\n");\n' +                        
...     '}'             
'class B {\n' +                                          
  '  public:\n' +                                        
  '    explicit B(const A&amp; a) : a_{a} {}\n' +         
  '    const A&amp; a_;\n' +
  '};\n' +                                                
  '\n' +                                                  
  'int main() {\n' +                                      
  '  A a; B b{a};\n' +                                    
  '  printf("HERE\\n");\n' +                              
  '}'                                                     
> hljs.highlight("cpp", str)                              
{                                                         
  illegal: true,                                          
  relevance: 0,                                           
  value: 'class B {\n' +                                  
    '  public:\n' +                                       
    '    explicit B(const A&amp;amp; a) : a_{a} {}\n' +   
    '    const A&amp;amp; a_;\n' +                        
    '};\n' +                                              
    '\n' +                                                
    'int main() {\n' +                                    
    '  A a; B b{a};\n' +                                  
    '  printf("HERE\\n");\n' +                            
    '}'                                                   
}                                 
```

After commenting out the line, it switches to per-line highlighting and works.

ppwwyyxx:
Given that `highlight.js` is not very robust (it failed for my above example), I think a good strategy is to:
1. try highlight the whole block
2. if failed, highlight each line one by one
3. if failed, return original text

What do you think?

Still, I think the if-statement should be removed because there is no reason to use different logic to highlight code.

ppwwyyxx:
Or, an alternative to make highlighting more robust is to use `highlight(..., ignore_illegals=true)`. Because according to https://highlightjs.readthedocs.io/en/latest/api.html#library-api, line-by-line highlighting (as is done now in `tryHighlight`) is not robust:

> Note: continuation is NOT intended to support line-by-line highlighting because there is no requirement that a grammar handle linebreaks in any special way. It’s quite possible for a grammar to have a single mode/regex that matches MANY lines at once. This is not discouraged and entirely up to the grammar.

curbengh:
> I don't know the reason why this if() statement was introduced.
So I cannot decide to remove this if() statement.

It's introduced along with hljs option https://github.com/hexojs/hexo-util/pull/30. I don't know why either.

I tested this PR with/without line numbering, looks the same.

ppwwyyxx:
Any updates?
I think the if-statement is introduced for no reason in the first place. As a result it does lead to different behaviors depending on `options.hljs`.

seaoak:
I'm sorry, I'm too late.

I think @ppwwyyxx 's proposal will be better than current implementation.

That is,

  1. try highlight the whole block
  2. if failed, try `highlight(..., ignore_illegals=true)` for the whole block
  3. if failed, highlight each line one by one
  4. if failed, return original text (with the HTML class name `plain`)

Maybe the 2nd choice never fail.
But the document of the API does not say it clearly.
So the 3rd and the 4th choices will remain as current implementation.

In any case, the language name should be set as an HTML class name.
(see Issue #164)

SukkaW:
@seaoak @ppwwyyxx LGTM then.

PR #172, opened 2020-01-04
feat(prism): add mark & firstLine option support

SukkaW:
Related documents:

- `firstLine`: https://prismjs.com/plugins/line-numbers/
- `mark`: https://prismjs.com/plugins/line-highlight/

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,7 +453,7 @@ Option | Description | Default
 
 When `isPreprocess` is enabled, `prismHighlight()` will return PrismJS processed HTML snippet. Otherwise `str` will only be escaped and `prismHighlight()` will return the HTML snippet that is suitable for `prism.js` working in the Browser.
 
-`mark` options will have effect only when `isPreprocess` is disabled. `firstLine` options will have effect only when `isPreprocess` is disabled and `lineNumber` is enabled.
+`mark` and `firstLine` options will have effect only when `isPreprocess` is disabled.
 
 ### relative_url(from, to)
 
--- a/README.md
+++ b/README.md
@@ -448,8 +448,12 @@ Option | Description | Default
 `lang` | Language | `'none'`
 `tab`| Replace tabs |
 `isPreprocess` | Enable preprocess or not | true
+`mark` | Highlight specific line |
+`firstLine` | First line number |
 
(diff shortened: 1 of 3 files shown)


seaoak:
I have a question.

Currently, `firstLine` option will have effect only when `lineNumbers=true`.
This behavior describes clearly in `README`.

But I think some users might want to highlight lines with `lineNumbers=false`.

For example, with an options:

```js
options = {
  isPreprocess: false,
  lineNumbers: false,
  firstLine: 101,
  mark: '103',
}
```

the user might intend to highlight the 3rd line.

It is achieved by adding the attribute `data-line-offset="100"` for `pre` element.
https://prismjs.com/plugins/line-highlight/

```js
  if (!lineNumber && firstLine && !isPreprocess) preTagAttrArr.push(`data-line-offset="${firstLine-1}"`);
```

Is this feature useful or not ?

SukkaW:
@seaoak We should make sure that it won't break `backtick_filter` and `code()` helper with highlight.js.

seaoak:
@SukkaW I found that `highlight()` just behaves like above feature.

For example, with the options:

```js
options = {
  gutter: false,
  firstLine: 101,
  mark: '103',
}
```

`highlight()` marks the 3rd line.
That is, `firstLine` option has effect even if `gutter=false`.

Certainly, in `hexo:lib/plugins/filter/before_post_render/backtick_coode_block.js`,
`firstLine` option is set only when `gutter=true`.
But, as a result, above feature has no effect for `backtick_code_block` filter (not break).

And also, in `hexo:lib/plugins/tag/code.js`,
`firstLine` option is independent of `line_number` option (= `gutter` option).
For example, a snippet below highlights the 3rd line.

```plain
{% codeblock line_number:false first_line:101 mark:103 %}
```

I think this feature will increase compatibility with `highlight.js`.
Could you consider this feature ?

SukkaW:
According to PrismJS documents:

> In case you want the line numbering to be offset by a certain number (for example, you want the 1st line to be number 41 instead of 1, which is an offset of 40), you can additionally use the data-line-offset attribute.

@seaoak 

So we should accept `firstLine` no matter `lineNumbers` enabled or not, and apply offset as long as `firstLine` is given?

seaoak:
@SukkaW I agreed.

Could you add test case for generating the attribute `data-line-offset` ?

SukkaW:
@seaoak Related test case added.

Issue #174, opened 2020-01-05
toc helper should not parse the text of children element(s)


noraj:
## Check List

Please check followings before submitting a new feature request.

- [x] I have already read [Docs page](https://hexo.io/docs/)
- [x] I have already searched existing issues

## Feature Request

Fix the toc helper to select only text content and not other sub-elements like links.

In depth explanation here: https://github.com/hexojs/hexo-renderer-markdown-it/issues/102

SukkaW:
@curbengh 

```html
<h1 id="title-3"><span>Title</span> 3</h3>
```

In this case `Title 3` is what we needed for `text`.

curbengh:
@SukkaW 

Since the output is `'\n#\nCreate a new post\n'`, what if tocObj trim the newlines and anything before and after them? It maybe more appropriate to trim in toc helper, since user might want to retain the original (rare, but possible).

SukkaW:
@curbengh 

```html
<h3 id="create-a-new-post">
<a class="header-anchor" href="#create-a-new-post">#</a>Create a new post
</h3>
```

will become `\n#Create a new post\n`. Still not appropriate.

curbengh:
Looks like a common question in cheerio/jquery, 

- https://stackoverflow.com/questions/20832910/get-text-in-parent-without-children-using-cheerio
- https://www.designcise.com/web/tutorial/how-to-get-html-elements-text-excluding-children-using-jquery-and-javascript

Seems possible to fix, if htmlparser has the equivalent function of native DOM's `document.childNodes` or cheerio's `$.first()`.

Microbai:
When contents likes this 
```html
<h2><span id="title_1">title 1</span></h2>
```
 tocObj won't parse the correct id, So I edit the getId function, then work.
```js
const getId = ele => {
  const { id } = ele.attribs;
  const { children } = ele;
  return id || (!children[0] ? null : getId(children[0]));
};
```

SukkaW:
@Microbai `tocObj()` is not designed to get id from child element but only from parent element.

Microbai:
@SukkaW I Knew it.But when I use Hexo 4.2, the page can't render toc correctly.So I edit tocObj for now, and it works.

PR #175, opened 2020-01-13
fix(tocObj): skip permalink symbol 

curbengh:
Fixes https://github.com/hexojs/hexo-util/issues/174 Closes https://github.com/hexojs/hexo-renderer-markdown-it/issues/102

Permalink (in this context) is a single non-word character, word = [a-Z0-9], and wrapped in `<a>`.
It may be wrapped in whitespace(s)

Scenario | Input | Output (text) | Explanation
--- | --- | --- | ---
1 | `<h1 id="foo"><a>#</a>bar</h1>`<br><br>`<h1 id="foo">bar<a>#</a></h1>`<br><br>`<h1 id="foo"><a>#</a>bar<a>#</a></h1>` | `bar`<br><br>`bar`<br><br>`bar` | `#` is treated as a permalink and skipped
2 | `<h1>号码<a>牌</a></h1>` | `号码` | `牌` is treated as a permalink
3 | `<h1><a>#</a></h1>` | `#` | If there is only permalink, parse it
4 | `<h1><a>b</a> two</h1>` | `b two` | `b` is not a permalink

---

returns empty string `''` instead of null. This is following convention. htmlparser2 `DomUtils.getText(elNoText)` and [camaro](https://www.npmjs.com/package/camaro) `camaro.transform(xml, {foo:'invalid'})` also return empty string. This is [compatible](https://github.com/hexojs/hexo/blob/79bb3aab3bef68d55798a2fe2eed2496dc8562dc/lib/plugins/helper/toc.js#L28) with toc, as empty string is still falsy (same as null and undefined).

### DIFF
--- a/test/toc_obj.spec.js
+++ b/test/toc_obj.spec.js
@@ -87,4 +87,73 @@ describe('tocObj', () => {
 
     result.length.should.eql(0);
   });
+
+  it('empty text', () => {
+    const input = '<h1></h1>';
+    const result = tocObj(input);
+
+    result[0].text.should.eql('');
+  });
+
+  describe('children element', () => {
+    it('<a> element with permalink + text', () => {
+      const input = [
+        '<h1><a>#</a>foo</h1>',
+        '<h1>foo<a>#</a></h1>',
+        '<h1><a>#</a>foo<a>#</a></h1>',
+        '<h1><a># </a>foo</h1>',
+        '<h1><a># </a>foo<a> #</a></h1>',
+        '<h1><a>号</a>foo</h1>'
+      ];
+      const result = input.map(str => tocObj(str));
+
+      result.forEach(str => str[0].text.should.eql('foo'));
+    });
+
+    it('<a> element - no text', () => {
+      const input = '<h1><a>foo</a></h1>';
+      const result = tocObj(input);
+
+      result[0].text.should.eql('foo');
+    });
+
+    it('non-permalink <a> element + text', () => {
+      const input = [
+        '<h1><a>foo</a>bar</h1>',
+        '<h1>foo<a>bar</a></h1>'
+      ];
+      const result = input.map(str => tocObj(str));
+
+      result.forEach(str => str[0].text.should.eql('foobar'));
+    });
+
+    it('non-permalink <a> element + unicode text', () => {
+      const input = [
+        '<h1><a>这是</a>测试</h1>',
+        '<h1>这是<a>测试</a></h1>'
+      ];
+      const result = input.map(str => tocObj(str));
+
+      result.forEach(str => str[0].text.should.eql('这是测试'));
+    });
+
+    it('multiple <a> elements', () => {
+      const input = '<h1><a>foo</a><a>bar</a></h1>';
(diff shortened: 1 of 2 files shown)


test/toc_obj.spec.js
```diff
@@ -87,4 +83,38 @@ describe('tocObj', () => {
...
+
+    result[0].text.should.eql('');
+  });
+
+  describe('children element', () => {
+    it('children + text', () => {
+      const input = [
+        '<h1 id="foo"><a>#</a>bar</h1>',
+        '<h1 id="foo">bar<a>#</a></h1>',
+        '<h1 id="foo"><a>#</a>bar<a>#</a></h1>'
```
SukkaW:
```suggestion
        '<h1 id="foo"><a>#</a>bar<a>#</a></h1>',
        '<h1 id="foo"><a>#</a><span>bar</span><a>#</a></h1>'
```

curbengh: `<h1 id="foo"><a>#</a><span>bar</span><a>#</a></h1>` would result in `#bar#` though since `<span>bar</bar>` is an element.

curbengh: A workaround is `if (element.type === 'text' || element.type === 'span')`.

`<h1><a>#</a><span>bar</span><a>#</a></h1>` -> `bar`

But the workaround is not foolproof,

`<h1><a>#</a>hey<span>bar</span><a>#</a></h1>` -> `hey`
`<h1><a>#</a><span>bar</span>hey<a>#</a></h1>` -> `bar`

Since `for..of` doesn't follow order, the result can be erratic.

SukkaW: The children tag could be anything: `<span>`, `<a>` even `<p>`.

curbengh:
`<h2>foo<i>bar</i>baz</h2>` only results in `foo`. I think tocObj should just avoid `<a>`.

SukkaW:
> `<h2>foo<i>bar</i>baz</h2>` only results in `foo`. I think tocObj should just avoid `<a>`.

So what happened if `<h2><a href="">foobar</a></h2>` or `<h2><a>foo</a> bar</h2>`?

curbengh:
with the new workaround (will commit in a minute),

`<h2><a href="">foobar</a></h2>` -> `foobar`
`<h2><a>foo</a> bar</h2>` -> ` bar` (foo is assumed to be a permalink symbol)

I think 2nd case is rare, if a user wants to add a link, 1st case would be used.

SukkaW:
@curbengh I have a better idea. Using `replace` to remove the symbol after `getText`. The symbol should be passed as a parameter.

curbengh:
> The symbol should be passed as a parameter

parameter of which function?

curbengh:
rebased to include https://github.com/hexojs/hexo-util/pull/181.

noraj:
@curbengh Awesome, this will be released in 1.9.0? 1.8.2?

curbengh:
should be part of [1.9.0](https://github.com/hexojs/hexo-util/pull/182)

PR #176, opened 2020-01-18
htmlTag - toString() before match() for scalar values.

KentarouTakeda:
Fixed crash when entering a numerical value into `fb_app_id` of` open_graph` helper.

Facebook app_id is a numeric string, but if it is set in `_config.yml` without quotes, the yml parser will recognize it as number.

1.8.0 and later versions crashed if the value was set to number.

### DIFF
--- a/lib/html_tag.js
+++ b/lib/html_tag.js
@@ -25,7 +25,7 @@ function htmlTag(tag, attrs, text, escape = true) {
     if (attrs[i] == null) result += '';
     else {
       if (i.match(regexUrl)
-        || (tag === 'meta' && !attrs[i].toString().match(regexMeta) && Object.values(attrs)[0].match(regexMeta))) {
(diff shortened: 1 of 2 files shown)


lib/html_tag.js
```diff
@@ -25,7 +25,7 @@ function htmlTag(tag, attrs, text, escape = true) {
     if (attrs[i] == null) result += '';
     else {
       if (i.match(regexUrl)
-        || (tag === 'meta' && !attrs[i].match(regexMeta) && Object.values(attrs)[0].match(regexMeta))) {
+        || (tag === 'meta' && !attrs[i].toString().match(regexMeta) && Object.values(attrs)[0].match(regexMeta))) {
```
SukkaW:
```suggestion
        || (tag === 'meta' && String(!attrs[i]).match(regexMeta) && Object.values(attrs)[0].match(regexMeta))) {
```

I prefer `String()` since only Object and Number have `toString()`.

KentarouTakeda: That's right. Thanks!!

PR #177, opened 2020-01-26
feat(prism): add data-language attribute

SukkaW:
Add `data-language` attribute to `<pre>` tag.

It is necessary to have `data-language` attribute to be compatible with Prism's `Show Languages` Plugin: https://prismjs.com/plugins/show-language/

Besides that, with this attribute theme developer can even show current language without that plugin as well:

``` css
pre::before {
    color: #bcc3ce;
    content: attr(data-language); /* This looks important */
    font-size: .7rem;
    position: absolute;
    right: .4rem;
    top: .1rem;
}
```

will resulted in (The `HTML` string on top of the right):

image

### DIFF
--- a/lib/prism.js
+++ b/lib/prism.js
@@ -68,13 +68,14 @@ function PrismUtil(str, options = {}) {
 
   const preTagClassArr = [];
   const preTagAttrArr = [];
+  let preTagAttr = '';
 
   if (lineNumber) preTagClassArr.push('line-numbers');
   preTagClassArr.push(`language-${language}`);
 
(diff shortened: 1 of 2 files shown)


seaoak:
When no `lang` is specified,
this patch add the attribute `data-language="plain"`.

Is this behavior intended by Prism's `show-language` plugin?

Maybe adding no attribute be better?

SukkaW:
> this patch add the attribute `data-language="plain"`.

Emmm, it will add `data-language="none"` instead. I will push a fix to handle this.

PR #181, opened 2020-02-06
refactor(toc_obj): simplify the code

SukkaW:


lib/toc_obj.js
```diff
@@ -8,28 +8,26 @@ const parseHtml = html => {
...
 
 function tocObj(str, options = {}) {
-  options = Object.assign({
+  const { min_depth, max_depth } = Object.assign({
     min_depth: 1,
     max_depth: 6
   }, options);
 
-  const headingsSelector = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'].slice(options.min_depth - 1, options.max_depth).join(',');
+  const headingsSelector = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'].slice(min_depth - 1, max_depth);
```
SukkaW: Array has `includes()` as well, remove unnecessary `.join()`.

lib/toc_obj.js
```diff
@@ -8,28 +8,26 @@ const parseHtml = html => {
...
-  const dom = parseHtml(str);
-  const headings = DomUtils.find(el => headingsSelector.includes(el.tagName), dom, true);
+  if (!headingsLen) return [];
 
   const result = [];
 
-  if (!headings.length) return result;
-
-  for (const el of headings) {
+  for (let i = 0; i < headingsLen; i++) {
```
SukkaW: Use classic loop.

lib/toc_obj.js
curbengh: Array also works better since the value must be exact match.
Currently headingsSelector also matches `h` tag.

PR #183, opened 2020-02-16
fix(is_external_link): handle invalid url

SukkaW:
Details can be found at https://github.com/hexojs/hexo/issues/4130

This PR introduce a try block and treat invalid URL as internal link (return `false`).

### DIFF
--- a/lib/is_external_link.js
+++ b/lib/is_external_link.js
@@ -13,6 +13,9 @@ const cache = new Cache();
 
 function isExternalLink(input, sitehost, exclude) {
   return cache.apply(`${input}-${sitehost}-${exclude}`, () => {
+    // Return false early for internal link
+    if (!/^(\/\/|http(s)?:)/.test(input)) return false;
+
     sitehost = parse(sitehost).hostname || sitehost;
 
     if (!sitehost) return false;
--- a/lib/is_external_link.js
+++ b/lib/is_external_link.js
@@ -16,8 +16,15 @@ function isExternalLink(input, sitehost, exclude) {
     sitehost = parse(sitehost).hostname || sitehost;
 
     if (!sitehost) return false;
-    // handle relative url
-    const data = new URL(input, `http://${sitehost}`);
+
+    // handle relative url and invalid url
+    let data;
+    try {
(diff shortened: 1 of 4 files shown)


SukkaW:
According to [travis ci build](https://travis-ci.com/sukkaw/hexo/builds/149168430), no performance impact detected.

SukkaW:
`is-absolute-url` is also added for better performance.

Since it is a widely used packages on npm so I believe it is reliable enough.

lib/is_external_link.js
```diff
@@ -13,11 +14,19 @@ const cache = new Cache();
 
 function isExternalLink(input, sitehost, exclude) {
   return cache.apply(`${input}-${sitehost}-${exclude}`, () => {
-    sitehost = parse(sitehost).hostname || sitehost;
+    if (!isAbsoluteUrl(input)) return false;
```
curbengh: `isAbsoluteUrl` matches [any](https://github.com/sindresorhus/is-absolute-url/blob/8a57ccec064d925640860674cd32e7182eeb3596/index.js#L15) protocol. For our use case, `/^http(s)?:\/\//.test(input)` alone should be sufficient.

SukkaW: @curbengh Agreed. I will update to it then.

SukkaW: Updated to handle `http://`, `https://` and **`//`** prefixed url.

lib/is_external_link.js
```diff
@@ -13,11 +13,21 @@ const cache = new Cache();
...
 
     if (!sitehost) return false;
-    // handle relative url
-    const data = new URL(input, `http://${sitehost}`);
+
+    // handle relative url and invalid url
+    let data;
+    try {
+      data = new URL(input, `http://${sitehost}`);
+    } catch (e) { }
```
jiangtj: How about `hexo.log.warn(...)` to alert the user that there has an invalid URL?

SukkaW: Since hexo-util has no `hexo-log`'s dependency, there is no way we can use `hexo.log.warn` in hexo-util.

PR #191, opened 2020-04-21
chore: drop Node.js 8

SukkaW:
https://github.com/hexojs/hexo/issues/3818


curbengh:
lgtm, just need to rebase.

SukkaW:
@curbengh Done.

Minimum required Node.js version is set to 10.3.0 (Tested on my local machine).

curbengh:
> Minimum required Node.js version is set to 10.3.0

can increase to at least 10.9 just in case we need 10.9+ features, prefer 10.18 if possible.

SukkaW:
> prefer >=10.18 if possible.

@curbengh As I mentioned in https://github.com/hexojs/hexo/issues/3818#issuecomment-619317156, we could use 10.13.0 instead (which is the first LTS version of Node.js 10)

PR #195, opened 2020-04-24
fix: avoid overriding Transform.destroy() method

curbengh:
fix compatibility with Node 14

https://github.com/hexojs/hexo/issues/4260#issuecomment-618483486

### DIFF
--- a/lib/cache_stream.js
+++ b/lib/cache_stream.js
@@ -18,10 +18,6 @@ CacheStream.prototype._transform = function(chunk, enc, callback) {
   callback();
 };
 
-CacheStream.prototype.destroy = function() {
-  this._cache.length = 0;
-};
-
 CacheStream.prototype.getCache = function() {
   return Buffer.concat(this._cache);
 };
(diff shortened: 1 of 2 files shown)


curbengh:
Introduce regression in Node < 14.

curbengh:
Tested Node 10, 12, 13 and 14 on dummy blog using `hexo g --debug`, no errors and resolved empty files issue.

stevenjoezhang:
Confirm that it is working properly now.
However, after removing the definition of `destroy` method, `cacheStream._cache` will not be deleted, which means that the resource is actually processed by JavaScript garbage collection, not `cacheStream.destroy();` here:

https://github.com/hexojs/hexo/blob/b60f815b975178ee1c907a92e5a1403830e56f37/lib/plugins/console/generate.js#L68

SukkaW:
> which means that the resource is actually processed by JavaScript garbage collection

Will it cause memory leak?

SukkaW:
Verified in https://github.com/hexojs/hexo/pull/4264

curbengh:
> cacheStream._cache will not be deleted...the resource is actually processed by JavaScript garbage collection

Meaning `destroy()` triggers the garbage collection to clean up `_cache`? Will `_cache` get cleaned up and not cause memory leakage?

SukkaW:
> I assume `destroy()` is working then?

Yes. If it is not working, there might be huge memory consumption.

PR #196, opened 2020-04-25
docs(highlight): 'sublanguage highlight' requirement

curbengh:
breaking change introduced by highlight.js v10 https://github.com/hexojs/hexo-util/pull/192.


stevenjoezhang:
Relevant issue https://github.com/hexojs/hexo/issues/1786

curbengh:
I add a note that `autoDetect` is slow.

PR #198, opened 2020-04-27
feat(hash): drop HashStream()

curbengh:
this finally drops `HashStream()` which has been long deprecated since https://github.com/hexojs/hexo-util/pull/45.


SukkaW:
FYI, hexo is still relying on `HashStream`, which means should be dropped after hexo-util releasing a new version.

https://github.com/search?q=org%3Ahexojs+HashStream&type=Code

PR #212, opened 2020-06-12
feat(workerpool): bring up

SukkaW:
The PR is a part of hexojs/hexo#4355.

Bring up a utility to manage workers and job queue.


SukkaW:
Minimum required Node.js version has bumped to 12.13, which is the first LTS version of Node.js 12: https://nodejs.org/en/blog/release/v12.13.0/

scripts/dummy_worker.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+// eslint-disable-next-line node/no-unsupported-features/node-builtins
```
curbengh, line 3: not required after https://github.com/hexojs/hexo-util/pull/212/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R57
comment in other files can be removed too.

SukkaW:
@curbengh We might not bring up Worker Threads during Hexo 5.0.0 development. So we can leave this PR here until we officially drop Node.js 12.

stevenjoezhang:
Any updates on this thread?

PR #213, opened 2020-06-17
fix: escapeHTML cannot esacpe numeric character reference correctly

rqy1458814497:
The old implementation will escape '&' in any numeric character reference except 
```text
', `, /, =
```

For example if someone using some language with Unicode character, e.g.  Σ (`&#931;`) will escape as `&amp;#931;`, then display as `&#931;` not `Σ`

(First I thought it's a fault of theme-next in theme-next/hexo-theme-next#1521)

### DIFF
--- a/lib/escape_html.js
+++ b/lib/escape_html.js
@@ -19,7 +19,7 @@ function escapeHTML(str) {
   str = unescapeHTML(str);
 
   // http://stackoverflow.com/a/12034334
-  return str.replace(/[&<>"'`/=]/g, a => htmlEntityMap[a]);
+  return str.replace(/[<>"'`/=]/g, a => htmlEntityMap[a]).replace(/&(?!#(\d{1,4}|x[a-zA-Z0-9]{1,4});)/, '&amp;');
 }
 
 module.exports = escapeHTML;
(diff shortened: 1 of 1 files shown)


rqy1458814497:
actually i find that unescape then re-escape certainly cannot work for all case: it will still treat `&amp;#x931;` as same as `&#x931;`... unless we unescape all possible escaping-sequence... and this sounds effectiveness

rqy1458814497:
For example if I want to paste these comments to my blog (although I don't) then the code highlight must be wrong, I think

rqy1458814497:
> According to [theme-next/hexo-theme-next#1521](https://github.com/theme-next/hexo-theme-next/issues/1521)
> Please uninstall `hexo-katex`. It's a 3rd-party plugin and is now deprecated.
> There is no need to modify the `escapeHTML` method. Thanks.

Sorry for it (and no one tells me what to do after updating hexo and hexo-next)... I actually do not know what package to use now

stevenjoezhang:
You can use the build-in KaTeX engine of theme NexT: https://theme-next.js.org/docs/third-party-services/math-equations.html
Or try this plugin: https://www.npmjs.com/package/hexo-filter-mathjax

PR #215, opened 2020-06-23
test(highlight): fix test cases

SukkaW:
Fix failed test cases.


curbengh:
The quote should already be escaped by [`escapeHTML()`](https://github.com/hexojs/hexo-util/blob/c92c5e3db16045f73208a8129fe67d36c07b9021/lib/escape_html.js#L9), I'm guessing previously highlight.js unescape it?

PR #216, opened 2020-06-26
feat(highlight): add prism-themes

stevenjoezhang:
`prism-themes` lists a selection of additional themes for the Prism syntax highlighting library: https://github.com/PrismJS/prism-themes/tree/master/themes
Including this package in `dependencies` will make it easier for theme developers to use code highlight style from `node_modules`. e.g.
```scss
@import '../node_modules/highlight.js/styles/a11y-dark.css';
@import '../node_modules/prismjs/themes/prism-coy.css';
@import '../node_modules/prism-themes/themes/prism-a11y-dark.css';
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "deepmerge": "^4.2.2",
     "highlight.js": "^10.0.0",
     "htmlparser2": "^4.0.0",
+    "prism-themes": "^1.4.0",
     "prismjs": "^1.17.1",
     "strip-indent": "^3.0.0",
     "striptags": "^3.1.1"


SukkaW:
I would rather not to do that.

Using other package manager instead of `npm` (likes `yarn`, `pnpm`) will result in different `node_modules` structure. So theme developers are still required to resolve the path themselves.

IMHO, `prism-theme` should only be the dependencies of the theme, rather than `hexo-util`'s.

stevenjoezhang:
Indeed, this Pull Request does not reduce the work of theme developers. Developers still need to use some code to find the installation location of prism and highlight.js.
However, this is more user-friendly. At present, most themes are installed via `git clone` instead of `npm install`. If `prism-theme` is a dependency of the Hexo theme, then the user needs to execute an additional `npm install` command in the theme directory. 
Therefore, I believe it is more appropriate to make it a dependency of hexo-util until one day "installing the theme through npm" becomes mainstream. Using a single dependency package to enhance the experience of all prismjs users, this is undoubtedly acceptable.

SukkaW:
@stevenjoezhang But theme developer still have to deal with different `node_modules` structures.

Unlike JS, CSS (Sass, Less, Styl) has no `requires.resolve` or sorts of things. (Not mentioning `yarn@2` even has no `node_modules` any more).

stevenjoezhang:
It is possible to get the root directory of prism with some tricks. e.g.
```js
const fs = require('fs');
const path = require('path');

let dir = path.dirname(require.resolve('prism-themes/package.json'));
hexo.theme.config.prism_style = `${dir}/themes/prism-a11y-dark.css`;
```
```scss
@import hexo-config('prism_style');
```

stevenjoezhang:
> Not mentioning `yarn@2` even has no `node_modules` any more

`hexo.plugin_dir` is `node_modules`, does it need to be rewritten?
https://github.com/hexojs/hexo/blob/f4f551095dd92459b81d77d0530b40f58ef56585/lib/hexo/index.js#L95
https://github.com/hexojs/hexo/blob/f4f551095dd92459b81d77d0530b40f58ef56585/lib/hexo/load_config.js#L44

SukkaW:
@stevenjoezhang 

Yarn 2 still supports `node_modules` but requires manually enables it:

```yaml
# .yarnrc.tml
nodeLinker: true
```

We could mention it in troubleshooting.

stevenjoezhang:
Yes, and `hexo-cli` is using `resolve` (https://github.com/browserify/resolve) instead of `require.resolve`
It also needs to be configured, otherwise `hexo-cli` cannot find the installation location of `hexo`

https://github.com/hexojs/hexo-cli/blob/2b07acbef9930a899456a15faecb4113b1c58308/lib/hexo.js#L85

stevenjoezhang:
Never mind. I have updated the docs of NexT.

PR #217, opened 2020-07-01
fix(full_url_for): handle config.url with a trailing slash

curbengh:
a notable example is Github Pages with project page, e.g. https://github.com/OpenTechSchool/social-coding corresponds to https://opentechschool.github.io/social-coding/. Project page always have a trailing slash, accessing `https://opentechschool.github.io/social-coding` will 301-redirect to `https://opentechschool.github.io/social-coding/`. So, in this case, user may prefer to use `url: https://opentechschool.github.io/social-coding/`.

This PR adds support for that.


stevenjoezhang:
Yes, and there is a hint in the `_config.yml`
> If your site is put in a subdirectory, set url as 'http://yoursite.com/child' and root as '/child/'

https://github.com/hexojs/hexo-starter/blob/259ecc3982e0445075f6c0daaccc24e4f1c7e663/_config.yml#L15

Which means
```
url: https://example.com/blog/
```
should be considered as invalid config.

See also https://github.com/hexojs/hexo/issues/1812#issuecomment-213667682

lib/full_url_for.js
```diff
@@ -24,7 +24,7 @@ function fullUrlForHelper(path = '/') {
     // Exit if input is an external link or a data url
     if (data.hostname !== sitehost || data.origin === 'null') return path;
 
-    path = encodeURL(config.url + `/${path}`.replace(/\/{2,}/g, '/'));
+    path = encodeURL((config.url + `/${path}`).replace(/\/{2,}/g, '/'));
```
jiangtj, line 27: I printed the log here, although `encodeURL` make the url right but I don't think this is the right change.

```js
    const ddd = (config.url + `/${path}`).replace(/\/{2,}/g, '/');
    console.log(ddd);
    path = encodeURL(ddd);
    console.log(path);
```

Output:
```
PS C:\Users\MrTT\Documents\GitHub\hexo-util> yarn test -f 'internal url'
yarn run v1.19.1
$ mocha -f "internal url"


  full_url_for
https:/example.com/index.html
https://example.com/index.html
https:/example.com/
https://example.com/
    √ internal url - root directory
https:/example.com/blog/index.html
https://example.com/blog/index.html
https:/example.com/blog/
https://example.com/blog/
    √ internal url - subdirectory
https:/example.com/blog/index.html
https://example.com/blog/index.html
https:/example.com/blog/
https://example.com/blog/
    √ internal url - subdirectory with trailing slash
https:/example.com/index.html
https://example.com/index.html
    √ internal url - no duplicate slash
```

curbengh:
> config.url = config.url.replace(/\/+$/, '');

that failsafe is not available in Hexo API (e.g. when use in a plugin's unit test).

SukkaW:
@curbengh 

`hexo.config` is added by `load_config`, thus the failsafe is always there. Plugins‘ unit test? They should be written in valid config format (because we already handle it). In case you are worrying about mocking test cases, use `await hexo.init() && await hexo.load()` will be fine then.

Also, current behavior of `full_url_for('/')` will always resulted in `hexo.config.url + '/'` (see `full_url_for`'s test cases). There is no need to worry about `trailing slash` issue.

https://github.com/hexojs/hexo-util/blob/14fa817f350cc4c318a411fbd52c0f0cce7cea2e/test/full_url_for.spec.js#L13-L15

Last but not least, all browsers, http request libs as well as crawlers will treat `//example.com` as `//example.com/`, and no 301 will be returned:

```http
GET https://blog.skk.moe

> HTTP/1.1 200 OK
```

```http
GET https://blog.skk.moe.

> HTTP/1.1 200 OK
```

```http
GET https://blog.skk.moe/

> HTTP/1.1 200 OK
```

curbengh:
> no 301 will be returned

I was referring to `url: http://example.com/blog/` example. `load_config.js` removes the trailing slash, so requesting `http://example.com/blog` will be 301-ed to `http://example.com/blog/`. [`feed`](https://github.com/hexojs/hexo-generator-feed/blob/fa1f04a9fac617f406eab8976af0e7883b7297b7/lib/generator.js#L46) currently does add it back, but for unrelated purpose (it was for [`config.url + path`](https://github.com/hexojs/hexo-generator-feed/pull/149). This remove/add trailing slash operation seems un-intuitive to me.

---

``` yml
## If your site is put in a subdirectory, set url as 'http://yoursite.com/child' and root as '/child/'
url: http://yoursite.com
root: /
```

My understanding of the rational behind `http://yoursite.com/child` limitation is that Hexo previously used `config.url + path`. After the availability of `full_url_for()`, that concat approach is no longer used (AFAIK) and we even discourage theme/plugin devs from using it. So, the limitation is no longer needed, as in it's now trivial to be flexible to existence of trailing slash.

In GH Pages, since `http://yoursite.com/child/` is the actual URL, it's natural for newcomers to use that; it's only _after_ they notice `## If your site is put in a subdirectory...` that the trailing slash needs to be removed.

Hexo blog can be hosted anywhere, we can't know whether `http://yoursite.com/child/` or `http://yoursite.com/child` is the actual one, a web server/hosting platform may prefer to have a trailing slash and vice versa. Since we can't know, why not let the user decides?

curbengh:
Fixed invalid protocol (`https:/example.com/index.html`) raised by @jiangtj 

---

`full_url_for('/')` now simply returns `config.url`:

```js
const hexo = {
  config: {
    url: 'http://example.com'
  }
}

console.log(full_url_for.call(hexo, '/'))
// http://example.com

hexo.config.url = 'http://example.com/'
console.log(full_url_for.call(hexo, '/'))
// http://example.com/

hexo.config.url = 'http://example.com/blog'
console.log(full_url_for.call(hexo, '/'))
// http://example.com/blog

hexo.config.url = 'http://example.com/blog/'
console.log(full_url_for.call(hexo, '/'))
// http://example.com/blog/
```

---

`full_url_for('index.html')` appends a trailing slash if `config.pretty_urls.trailing_index === false`:

```js
const hexo = {
  config: {
    url: 'http://example.com'
  }
}

console.log(full_url_for.call(hexo, 'index.html'))
// http://example.com/

hexo.config.url = 'http://example.com/'
console.log(full_url_for.call(hexo, 'index.html'))
// http://example.com/

hexo.config.url = 'http://example.com/blog'
console.log(full_url_for.call(hexo, 'index.html'))
// http://example.com/blog/

hexo.config.url = 'http://example.com/blog/'
console.log(full_url_for.call(hexo, 'index.html'))
// http://example.com/blog/
```

test/full_url_for.spec.js
```diff
@@ -12,13 +12,19 @@ describe('full_url_for', () => {
...
 
   it('internal url - subdirectory', () => {
     ctx.config.url = 'https://example.com/blog';
     fullUrlFor('index.html').should.eql(ctx.config.url + '/index.html');
-    fullUrlFor('/').should.eql(ctx.config.url + '/');
+    fullUrlFor('/').should.eql(ctx.config.url);
+  });
+
+  it('internal url - subdirectory with trailing slash', () => {
+    ctx.config.url = 'https://example.com/blog/';
```
SukkaW, line 25: No. The case here just won't happen.

https://github.com/hexojs/hexo/blob/17ee23fa181bcc2c3b47eba5b345762aef1eac36/test/scripts/hexo/load_config.js#L93-L106

Just as the test case of `load_config` shows, the `url: https://hexo.io/` will result in `hexo.config.url.should.eql('https://hexo.io');`, while `root: foo` will result in `hexo.config.root.should.eql('foo/');`. Thus, the test case just won't exists.

jiangtj:
In fact, I am confused about these two configurations.

```
url: http://yoursite.com  => root: /
url: http://yoursite.com/child  => root: /child/
```

If such a rule must meet, I think that only the configuration of `url` is enough, `root` can be generated during the initialization of hexo

Or is there a situation that does not meet this rule?

```
url: http://yoursite.com/child  ≠> root: /child/
```

test/full_url_for.spec.js
curbengh: https://github.com/hexojs/hexo/blob/17ee23fa181bcc2c3b47eba5b345762aef1eac36/lib/hexo/load_config.js#L34-L35 should be removed if this PR is accepted. Those two lines are considered a workaround and this PR handles it.

curbengh:
This PR makes it flexible to any combination of the following configs:

``` yml
url: http://yoursite.com/child
root: /child

url: http://yoursite.com/child/
root: /child

url: http://yoursite.com/child
root: /child/

url: http://yoursite.com/child/
root: /child/

url: http://yoursite.com/child/
root: child/

url: http://yoursite.com/child/
root: child

url: http://yoursite.com/child
root: child/

url: http://yoursite.com/child
root: child
```

---

@jiangtj 
> I think that only the configuration of url is enough, root can be generated during the initialization of hexo

You are right, it's trivial to derive `root` from `url`; we can make `root` config as optional and generate it if it's empty.

I wonder if there is any use case for:

``` yml
url: http://yoursite.com/
root: child
```

stevenjoezhang:
I'm not sure if this PR is being too lenient on invalid config.

curbengh:
> I'm not sure if this PR is being too lenient on invalid config.

This PR may break `{{ url + path }}` or `{{ root + path }}`, however, these approaches are discouraged and we recommend using `full_url_for(path)` or `url_for(path)` instead.

The purpose of this PR is reducing 301/302 redirects for anyone using `url: http://example.com/child` (possibly benefiting SEO) and encourage users to use ``url: http://example.com/child/` instead since this is the true URL most of the time.

The main goal is to make the config slightly more user-friendly.

SukkaW:
@jiangtj @curbengh @stevenjoezhang 

Please considering about Breaking Changes and the backward compatibility of current themes then.

If theme is designed with the current configuration format, then we shouldn't change the `root` & `url`.

jiangtj:
> You are right, it's trivial to derive `root` from `url`; we can make `root` config as optional and generate it if it's empty.
> 
> I wonder if there is any use case for:
> 
> ```yaml
> url: http://yoursite.com/
> root: /child/
> ```

@curbengh If there is no special case, I prefer to remove the `root` directly, and then generate it

```
url: http://yoursite.com  => {url: http://yoursite.com, root: /}
url: http://yoursite.com/  => {url: http://yoursite.com, root: /}
url: http://yoursite.com/child  => {url: http://yoursite.com/child, root: /child/}
url: http://yoursite.com/child/  => {url: http://yoursite.com/child, root: /child/}
```


> Please considering about Breaking Changes and the backward compatibility of current themes then.

@SukkaW If just make `full_url_for` more flexible, I think this is not BC. Because the documentation and many third-party plugin themes may use `url + path` or `root + path`, so there is no change for users. . .

curbengh:
> @SukkaW 
> If theme is designed to work with the current configuration format, then we shouldn't change the root & url in case things go wrong.
> @jiangtj 
> If just make full_url_for more flexible, I think this is not BC.

To clarify, this PR doesn't directly cause a BC; the flexibility I mentioned [earlier](https://github.com/hexojs/hexo-util/pull/217#issuecomment-657331828) is more of a _side effect_, not a primary intention of this PR, though I think it's a benefit. It is this flexibility that could be a BC.

However, I reckon all users with `root: /child/` uses `url: http://yoursite.com/child` as currently enforced by Hexo, so there is no immediate BC, `{{ url + path }}` still works.

With this PR, Hexo doesn't need (and shouldn't) enforce the syntax, so user can _then_ use:

``` yml
url: http://yoursite.com/child/
root: child
```

But that will break `{{ url + path }}`. Theme has two (easy) options:

1. `full_url_for()` (recommended)
2. User change config to `url: http://yoursite.com/child`

> @jiangtj 
> If there is no special case, I prefer to remove the root directly, and then generate it
> and both `full_url_for` and `url + path` `root + path` are valid

The intention is to keep the trailing slash `url: http://yoursite.com/child/`, so `{{ url + path }}` will not work without workaround.

jiangtj:
I mean when the user configures `url: http://yoursite.com/child/`, generate a new url without the trailing slash (`url: http://yoursite.com/child`) and the correct root (`root: /child/`)

So that `{{ url + path }}` can also work normally

PR #220, opened 2020-07-23
fix(spawn): support string argument

curbengh:
current example uses string argument, so `spawn()` should support it.
also add array argument and options example.


SukkaW:
BTW, would you mind release a newer version of hexo-util, including all latest changes, for Hexo 5.0.0 release?

PR #227, opened 2020-08-13
feat(prism): caption

curbengh:
Feature requested in https://github.com/hexojs/hexo/issues/4460 by @jerryc127

### DIFF
--- a/lib/prism.js
+++ b/lib/prism.js
@@ -62,7 +62,8 @@ function PrismUtil(str, options = {}) {
     tab,
     mark,
     firstLine,
-    isPreprocess = true
+    isPreprocess = true,
+    caption
   } = options;
 
   // To be consistent with highlight.js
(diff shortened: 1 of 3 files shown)


lib/prism.js
```diff
@@ -97,7 +98,9 @@ function PrismUtil(str, options = {}) {
 
   if (tab) str = replaceTabs(str, tab);
 
-  const startTag = `<pre class="${preTagClassArr.join(' ')}"${preTagAttr}><code class="language-${language}">`;
+  const figCaption = caption ? `<figcaption>${caption}</figcaption>` : '';
```
SukkaW, line 101: https://mdn.io/figcaption

`<figcaption>` is only valid in `<figure>` element.

SukkaW:
prismjs just doesn't support caption. Instead of adding caption support, what about simply ignoring it?

curbengh:
highlight.js doesn't support it either, but Hexo enhance the triple backtick to support it.

SukkaW:
So should we use other element instead? `<figcaption>` without `<figure>` is invalid.

curbengh:
`figcaption` is used for user's convenience so that existing `highlight.js` style can be easily adapted. I tried `<samp>` and `<kbd>`, the caption appears inline as codeblock.

samp

I haven't try all elements though.

SukkaW:
> A <figure> element; the <figcaption> element must be its first or last child.
> https://mdn.io/figcaption

`<kbd>` shouldn't be an option as well: most css already apply its own style: <kbd>kbd</kbd>

`<small>` is an inline element thus shouldn't be used.

curbengh:
[backtick_code_block.js](https://github.com/hexojs/hexo/blob/463025c9d21411570e56b15812eb319ea0883330/lib/plugins/filter/before_post_render/backtick_code_block.js#L36) already wraps in `<span>`,

``` js
// just wrap in <div>?
const codeCaption = caption ? `<div>${caption}</div> : '';
```

PR #228, opened 2020-08-15
feat(url_for): convert backward to forward slash

curbengh:
to deal with Windows path, see https://github.com/hexojs/hexo/pull/4479/files#r470968350

`url_for()` deals with relative URLs, which should never have backward slash.

`full_url_for()` doesn't need to update because `encodeURL()` already have conversion in place (via URL API).

### DIFF
--- a/lib/url_for.js
+++ b/lib/url_for.js
@@ -39,7 +39,7 @@ function urlForHelper(path = '/', options) {
     }
 
     // Prepend root path
-    path = encodeURL((root + path).replace(/\/{2,}/g, '/'));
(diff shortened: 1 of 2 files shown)


curbengh:
Alternatively, we could also convert in `encodeURL()` instead (used by both `url_for()` and `full_url_for()`.

SukkaW:
`url_for` is designed to transform the URL path. It should not be used to convert file system path into URL.

curbengh:
`full_url_for()` currently converts slash whereas `url_for()` percent-encode it, this PR makes them behave similarly.

PR #229, opened 2020-08-15
fix(highlight): use <div> when wrap is disabled

curbengh:
- BREAKING CHANGE: `<figcaption>` is replaced with `<div>` when wrap is disabled
- https://github.com/hexojs/hexo-util/pull/227#issuecomment-673347477

Affects users of following config:

``` yml
# _config.yml
highlight:
  line_number: false
  wrap: false

# or
prismjs:
  enable: true
```

New rendered html:

``` html
<pre>
  <div class="caption">
    caption here
  </div>
  <code>codeblock</code>
<pre>
```

suggested css:

``` css
pre div.caption {
  font-size: 0.9em;
  color: #888;
}

pre div.caption a {
  float: right;
}
```

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -30,7 +30,6 @@ function highlightUtil(str, options = {}) {
   const before = useHljs ? `<pre><code class="${classNames}">` : '<pre>';
   const after = useHljs ? '</code></pre>' : '</pre>';
 
-  const figCaption = caption ? `<figcaption>${caption}</figcaption>` : '';
 
   const lines = data.value.split('\n');
   let numbers = '';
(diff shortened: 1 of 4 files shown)


lib/highlight.js
```diff
@@ -43,15 +42,22 @@ function highlightUtil(str, options = {}) {
     content += formatLine(line, Number(firstLine) + i, mark, options, wrap);
   }
 
+  let codeCaption = '';
+
+  if (caption) {
+    if (wrap) codeCaption = `<figcaption>${caption}</figcaption>`;
+    else codeCaption = `<div>${caption}</div>`;
```
SukkaW, line 49:
```suggestion
    codeCaption = wrap ? `<figcaption>${caption}</figcaption>` : `<div>${caption}</div>`;
```

lib/highlight.js
```diff
@@ -43,15 +42,21 @@ function highlightUtil(str, options = {}) {
     content += formatLine(line, Number(firstLine) + i, mark, options, wrap);
   }
 
+  let codeCaption = '';
+
+  if (caption) {
+    codeCaption = wrap ? `<figcaption>${caption}</figcaption>` : `<div>${caption}</div>`;
```
SukkaW, line 48:
```suggestion
    codeCaption = wrap ? `<figcaption>${caption}</figcaption>` : `<div class="caption">${caption}</div>`;
```

For better css matching.

curbengh: updated `highlight` and `prismjs`.

PR #230, opened 2020-08-16
feat: joinPath()

curbengh:
ref: https://github.com/hexojs/hexo/pull/4479

this enhances [`path.join()`](https://nodejs.org/api/path.html#path_path_join_paths) to convert slash, but resolving '../' is not supported (e.g. `join('foo/bar/', '../baz')`) though.


README.md
```diff
@@ -324,6 +325,51 @@ htmlTag('script', {src: '/foo.js', async: true}, '')
...
+
+joinPath('foo/bar/baz', 'lorem\\ipsum\\dolor', { sep: '/' })
+// foo/bar/baz/lorem/ipsum/dolor
+
+joinPath('foo/', '/bar', '/baz')
+// Windows: foo\bar\baz
+// Unix: foo/bar/baz
+
+/* If the joined path string is a zero-length string, then '.' will be returned,
+representing the current working directory. */
```
curbengh, line 353: Similar behaviour as [`path.join()`](https://nodejs.org/api/path.html#path_path_join_paths)

jiangtj:
> but resolving '../' is not supported

[path.resolve()](https://nodejs.org/api/path.html#path_path_resolve_paths) is work?

curbengh:
> path.resolve() is work?

I mean this `joinPath()` doesn't support resolving path yet, perhaps in future if there is a demand, currently it's out of scope.

SukkaW:
@curbengh 

Recently I have found a new package on npm: https://npm.im/url-join

Have a look?

stevenjoezhang:
See also https://github.com/hexojs/hexo/pull/5457
I believe there are many more pieces of code that could utilize this `joinPath` function.

uiolee:
I think it should be implemented based on path or any other module instead of string concatenation.

stevenjoezhang:
Yes, maybe `path.posix.join` will work for such cases.
Moreover, some other languages have more comprehensive libraries for path manipulation, rather than just simple string operations. I particularly like Rust's `PathBuf`, which is very elegant. I'm not sure if Node.js has a similar implementation.

PR #231, opened 2020-08-17
release: 2.4.0

curbengh:
https://github.com/hexojs/hexo-util/pull/227 https://github.com/hexojs/hexo-util/pull/229

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hexo-util",
-  "version": "2.3.0",
+  "version": "2.4.0",
   "description": "Utilities for Hexo.",
   "main": "lib/index",
   "scripts": {


SukkaW:
Should #229 being considered as Breaking Changes?

`hexojs/hexo` has no `package.json` and `dependencies` is not pinned. If Hexo user installs the latest `hexo-util` their theme might be broken.

curbengh:
> Should #229 being considered as Breaking Changes?

I believe I'm the only affected user 😅

SukkaW:
> caption in `{ wrap: false }` config was only introduced recently in [2.2.0](https://github.com/hexojs/hexo-util/releases/tag/2.2.0) and it's not mentioned in hexo.io yet, so the breaking change is fairly minimal.

However there might be users who already uses caption white recently changed to `wrap: false`. Then `caption` could "come from nowhere" and break the theme.

curbengh:
> ...already uses caption while recently changed to `wrap: false`,  then caption could "come from nowhere"

if they already use caption, shouldn't they expect caption to be displayed (as introduced in https://github.com/hexojs/hexo-util/pull/210)? Switching to `wrap: false` requires a significant change to the theme and hardly any theme supports it because it doesn't support line number.

It doesn't necessarily break the theme, more of a missing style. since caption in `wrap: false` was not supported previously, the only users who would style the caption are the ones who closely follow this repo, and if they already closely following, it should fairly trivial for them to update caption style after #229. #229 is indeed a breaking change for them, but the change is too minimal to justify bumping major.

curbengh:
https://github.com/hexojs/hexo-util/releases/tag/2.4.0
https://www.npmjs.com/package/hexo-util/v/2.4.0

PR #234, opened 2020-11-18
feat(cache): add keyStringify option

SukkaW:
With `keyStringify` option, we could use a plain object as a cache key, and invalidate the cache when the object is changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,6 @@
 Utilities for [Hexo].
 
 ## Table of contents
-
 - [Installation](#installation)
 - [Usage](#usage)
 - [Cache](#cache)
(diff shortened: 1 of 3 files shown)


jiangtj:
So how to get the value of cache?

Also, I don't think hash is necessarily the best option, and I'd like to see a serialization function for key if possible.

For example
```js
Cache.default.serialization.MD5 = key => createSha1Hash('md5').update(key).digest('hex');

cache.apply({ a: 1, b: 2 }, 123, { keySerialization: Cache.default.serialization.MD5 }).should.eql(123);
cache.apply({ a: 1, b: 2 }, 123, { keySerialization: key => JSON.stringify(key) }).should.eql(123);
```

SukkaW:
Superseded by memoize proposal.

PR #260, opened 2021-09-03
fix(highlight): highlight sublanguages with tab

stevenjoezhang:
Issue resolved: https://github.com/hexojs/hexo/issues/4726


SukkaW (approved):
LGTM!

PR #265, opened 2021-10-20
stripHTML should be more strict with non string

tomap:
And strip them :)


tomap:
Note that this is breaking change, that would require a major version bump + an update of the documentation

SukkaW (approved):

ready-research:
@SukkaW @tomap Please validate the huntr issue https://www.huntr.dev/bounties/6a6de9c2-d2d2-4a07-8fc9-2d2dd2ea09bd/ using `Mark as Valid` and also `Confirm fix`. Thanks,

PR #269, opened 2022-01-11
feat(toc_obj): Support unnumbered headings

Cerallin:
This PR adds a feature for unnumbered headings.

1. The user specifies `unnumbered_class_list` and passes it to tocObj.
2. Headings WITHOUT classes in `unnumbered_class_list` will be parsed as usual.
3. Headings WITH classes in `unnumbered_class_list` will be parsed with `unnumbered: true`.

As is provided in `test/toc_obj.spec.js`, I have made minimized modifications as much as possible.

P.S. Maybe `unnumbered_class_list` is too long?


lib/toc_obj.js
```diff
@@ -13,11 +14,26 @@ const getId = ({ attribs = {}, parent }) => {
   return attribs.id || (!parent ? '' : getId(parent));
 };
 
+/**
+ * Identify a heading that to be unnumbered or not.
+ *
+ * @param {Node} {attribs.class} string contains all classes of the heading.
```
SukkaW, line 20:
```suggestion
 * @param {typeof import('domhandler').Node} {attribs.class} string contains all classes of the heading.
```

lib/toc_obj.js
```diff
@@ -1,4 +1,5 @@
 'use strict';
+const { Node } = require('domhandler');
```
SukkaW, line 2:
```suggestion
```

SukkaW, line 2:
```suggestion
```

lib/toc_obj.js
SukkaW, line 20:
```suggestion
 * @param {typeof import('domhandler').Node} {attribs.class} string contains all classes of the heading.
```

SukkaW (changes requested):

lib/toc_obj.js
```diff
@@ -13,11 +13,26 @@ const getId = ({ attribs = {}, parent }) => {
...
+ *
+ * @param {Node} {attribs.class} string contains all classes of the heading.
+ * @param {Set<string>} searchClassSet classes to search.
+ * @returns bool
+ */
+const isUnnumbered = ({ attribs = { class: "" } }, searchClassSet) => {
+  let classList = attribs.class;
+  return classList &&
+    classList.split(" ").filter(item => searchClassSet.has(item)).length > 0;
+};
```
SukkaW, line 27: IMHO passing the `unnumbered` through a `className` is kinda bad. What about using a `data-*` attribute? Like `<h1 class="" data-toc-unnumbered="false">`

Cerallin: That's a good idea. I will make a change later.

yoshinorin (commented):
@Cerallin 
Would you please update README?

> https://github.com/hexojs/hexo-util/tree/ab5ad9dd906f865faeba846b92be9dcce8c27120#tocobjstr-options

P.S: no need force-push. we will squash-merge.

README.md
```diff
@@ -10,38 +10,42 @@ Utilities for [Hexo].
...
-- [slugize](#slugizestr-options)
-- [spawn](#spawncommand-args-options)
-- [stripHTML](#striphtmlstr)
-- [wordWrap](#wordwrapstr-options)
-- [tocObj](#tocobjstr-options)
-- [truncate](#truncatestr-options)
-- [unescapeHTML](#unescapehtmlstr)
-- [url_for](#url_forpath-option)
-- [bind(hexo)](#bindhexo)
+- [hexo-util](#hexo-util)
```
SukkaW: It seems like you are using a VSCode extension that generates ToC automatically. Please disable it temporarily for the changes.

Cerallin: I've solved this in b53f3fa

yoshinorin (changes requested):
@Cerallin 
Would you please fix lint?

yoshinorin:
@Cerallin 
Continues from https://github.com/hexojs/hexo/pull/4871#issuecomment-1019077946

Do you have any idea to use this option with other renderers? Currently, hexo-renderer-marked and hexo-renderer-markdown-it don't have an option to add custom attributes. They are the official renderer of hexo. IMHO, it is better to use this option with official renderers. 

If difficult to implement it, would you please update the docs that users need to customize (extends) the renderer by themselves to add an attributes.

I assume users will be confused "this option can be used in official renderers by default".

@SukkaW 
Do you have any suggestions?

Cerallin:
@yoshinorin

Cerallin:
@yoshinorin 
This is an plugin for markdown-it:  [markdown-it-attrs](https://github.com/arve0/markdown-it-attrs).

Cerallin:
Support for `markdown-it-attrs` is added via https://github.com/hexojs/hexo-renderer-markdown-it/pull/174.

SukkaW (approved):

PR #271, opened 2022-01-31
test: fix generate coverage report & test failure when not supported language with Prismjs

yoshinorin:
Add comment for review.

### DIFF
--- a/.nycrc.yml
+++ b/.nycrc.yml
@@ -1,0 +1,2 @@
+# https://github.com/istanbuljs/nyc#common-configuration-options
+all: true
--- a/test/prism.spec.js
+++ b/test/prism.spec.js
@@ -77,8 +77,8 @@ describe('prismHighlight', () => {
       '}'
     ].join('\n');
 
-    // Use language: 'plain' to simplify the test
-    const result = prismHighlight(input, { tab: '  ', lang: 'plain' });
+    // Use language: 'plain-text' for not supported language with Prism
+    const result = prismHighlight(input, { tab: '  ', lang: 'plain-text' });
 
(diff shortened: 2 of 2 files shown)


test/prism.spec.js
```diff
@@ -77,8 +77,8 @@ describe('prismHighlight', () => {
       '}'
     ].join('\n');
 
-    // Use language: 'plain' to simplify the test
-    const result = prismHighlight(input, { tab: '  ', lang: 'plain' });
+    // Use language: 'plain-text' for not supported language with Prism
+    const result = prismHighlight(input, { tab: '  ', lang: 'plain-text' });
```
yoshinorin, line 81: I assume this test case seems to test for not supported language with Prism. But, if we specify `lang: 'plain` the `prismHighlight` function will return `Prism.highlight` result.

So, we want to pass through [return escapeHTML(code);](https://github.com/hexojs/hexo-util/blob/837008b7dd8633aee2484e01d3b85e1e0cf66c26/lib/prism.js#L28) when this test, but current test code pass through [return Prism.highlight(code, Prism.languages[language], language);](https://github.com/hexojs/hexo-util/blob/837008b7dd8633aee2484e01d3b85e1e0cf66c26/lib/prism.js#L24).

I think it happened before, not recently.

yoshinorin:
Hmm... coverage report sent failure when node 14.x 🤔

.nycrc.yml
```diff
@@ -0,0 +1,2 @@
+# https://github.com/istanbuljs/nyc#common-configuration-options
+all: true
```
yoshinorin, line 2: The coverage report file (`lcov.info`) is empty if we do not specify `all : true`.

tomap (approved):

PR #287, opened 2022-11-02
fix: Typographical problem of too long lines after code highlighting

EduarteXD:
The original implementation would have placed a table to hold the code inside the 'figure' tag
截屏2022-11-03 上午12 22 02
I implemented a code type identifier by placing the ::after pseudo-element in this 'figure'
截屏2022-11-03 上午12 24 30
I set 'overflow: scroll' to the <figure> element, so if scrolling occurs, the pseudo-element I set will scroll with it
截屏2022-11-03 上午12 28 50
The solution is to wrap the table with a div that has 'overflow: scroll', where the language indicator is correctly positioned
截屏2022-11-03 上午1 01 57

### DIFF
--- a/lib/highlight.js
+++ b/lib/highlight.js
@@ -54,7 +54,7 @@ function highlightUtil(str, options = {}) {
     return `<pre>${codeCaption}<code class="${classNames}"${languageAttr && lang ? ` data-language="${lang}"` : ''}>${content}</code></pre>`;
   }
 
-  let result = `<figure class="highlight${data.language ? ` ${data.language}` : ''}"${languageAttr && lang ? ` data-language="${lang}"` : ''}>`;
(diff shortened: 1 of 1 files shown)


stevenjoezhang (changes requested):
Some Hexo themes will handle this issue on the front end via javascript. This change will break the style of these themes.

SukkaW:
> But we could do a breaking change and bump major

@tomap We should not bring breaking changes to the theme whenever we want. It will take time for the theme developers to catch up and migrate (also considering some of the themes are no longer maintained). If we introduce a breaking change that the end user can not migrate, they will stop upgrading to the latest version of Hexo, which is something we don't want to happen.

SukkaW (changes requested):
No. Your issue is definitely fixable through CSS. It doesn't require any output HTML changes from the `hexo-util`.

You shouldn't introduce breaking changes to Hexo only because you are not good at CSS.

PR #291, opened 2022-11-14
update main in package.json

stevenjoezhang:


package.json
```diff
@@ -16,7 +16,7 @@
     "prepare": "npm run build:highlight"
   },
   "directories": {
-    "lib": "./lib",
+    "lib": "./dist",
```
SukkaW, line 19: Shouldn't it be `{ "dist": "./dist" }` instead?

yoshinorin (approved):

PR #296, opened 2023-01-02
fix(prism): use correct type to pass test

liby:
https://github.com/hexojs/hexo-util/blob/master/lib/prism.ts#LL57-L65C2

```
interface Options {
  caption?: string;
  firstLine?: number;
  isPreprocess?: boolean;
  lang?: string;
  lineNumber?: boolean;
  mark?: string;
  tab?: string;
}
```


yoshinorin (approved):
Thank you ❤️

PR #307, opened 2023-05-13
fix(url_for): some improvement for `url_for`

dimaslanjaka:
## chore: some improvement for `url_for`
1. make `options` as optional with default value as `empty object`
    - sample undefined `options` - https://github.com/next-theme/hexo-theme-next/blob/acca4e83747b168e90815945f58f982b1d14eac5/layout/_layout.njk#L10
    - sample usage in plugin - https://github.com/dimaslanjaka/hexo-shortcodes/blob/c099f51c412d794136ff9064bf0d182ae17a1573/src/gist.ts#L42-L49
2. add JSDoc
    - add sample usage inside plugin/scripts
3. assign `options` fallback to an `empty object`
    - sometimes i saw syntax using `url_for` the `options` given as `null`, prevent useless error give `options` fallback to an `empty object`


dimaslanjaka:
@SukkaW review pls

package.json
```diff
@@ -6,7 +6,7 @@
   "types": "./dist/index.d.ts",
   "scripts": {
     "prepublishOnly": "npm install && npm run clean && npm run build",
-    "build": "tsc -b",
+    "build": "npm run build:highlight && tsc -b",
```
SukkaW, line 9: `build:highlight` doesn't need to build every time.

SukkaW (commented):
Everything else LGTM!

package.json
dimaslanjaka: oh okay, ill delete it.

dimaslanjaka: deleted https://github.com/dimaslanjaka/hexo-util/commit/8663c657d2a231845ddc05f6f59264ccf92e2fa6

lib/url_for.ts
```diff
@@ -5,43 +5,73 @@ import prettyUrls from './pretty_urls';
...
   if (options.relative) {
     return relative_url(this.path, path);
   }
 
   const { root } = config;
-  const prettyUrlsOptions = Object.assign({
-    trailing_index: true,
-    trailing_html: true
-  }, config.pretty_urls);
+  const prettyUrlsOptions = Object.assign(
```
stevenjoezhang, line 47: You can keep the original code style here

uiolee (dismissed):
lgtm

lib/url_for.ts
uiolee: Looks like he used prettier. If eslint doesn't fail, I guess it doesn't matter.

lib/url_for.ts
```diff
@@ -5,43 +5,73 @@ import prettyUrls from './pretty_urls';
...
 const cache = new Cache<string>();
 
-function urlForHelper(path = '/', options) {
+/**
+ * url_for options type
+ * @example
+ * // to call this type
+ * type urlOpt = Parameters<typeof import('hexo-util')['url_for']>[1];
+ */
+interface url_for_options {
```
SukkaW, line 14:
```suggestion
interface UrlForOptions {
```

uiolee (approved):

PR #308, opened 2023-05-14
fix(Cache): typescript `Map` generic type fixed

dimaslanjaka:
## feat: add CacheMapper
implement built-in javascript `Map` with full function usages.
- Create new class `CacheMapper` implementing `Map<KEY, VALUE>` which is genericable
- Expose all the functionality of the Map class
- Add private inner Map<number, Foo> to prevent variable class imported using `private` name
- class `Cache` implementing `CacheMapper`

## fix: generic type `Cache` class
because `any` is not allowed by ESLint so I created a mapper class with full typestrong which can be generalized

## chore(deps-dev): add `@types/chai` and `@types/mocha`
to support mocha, chai for used in typescript, we need `@types/chai` and `@types/mocha`

## feat: add strict type test for `Cache`
define cache class with various type


dimaslanjaka:
@SukkaW can i add typescript test for generic type of `Cache` ?

i want add more tests within various type which `Cache` fully supported generic type

lib/CacheMapper.ts
```diff
@@ -0,0 +1,163 @@
+export class CacheMapper<K, V> implements Map<K, V> {
```
SukkaW, line 1: Can you make `V` to `any` by default? This would be helpful for the existing js code migration. We can always explicit define `T` anyway.

lib/CacheMapper.ts
```diff
@@ -0,0 +1,163 @@
...
+  private _innerMap: Map<K, V>;
+  size: number;
```
SukkaW, line 3: We should make `size` a read-only getter that returns `this._innerMap.size`.

lib/CacheMapper.ts
dimaslanjaka: yeah that was same as `T` because any generic type can be assigned to any `words without spaces`

dimaslanjaka: i create `K` for KEY and `V` for VALUE so can be easily undestand what that variable means

dimaslanjaka: and any generic constant are treat as `any` by default

dimaslanjaka: sorry for my bad english :)

dimaslanjaka: and we no need using `CacheMapper` class (we must hide this cache mapper from index export), we just using the `Cache<T>` class

test/cache.spec.js
```diff
@@ -1,63 +1,74 @@
...
 
 describe('Cache', () => {
+  // const Cache = require('../dist').Cache; // <-- this also works
   const Cache = require('../dist/cache');
   const cache = new Cache();
 
   it('get & set', () => {
     cache.set('foo', 123);
-    cache.get('foo').should.eql(123);
+    should.equal(cache.get('foo'), 123);
```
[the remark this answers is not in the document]
dimaslanjaka: > `@types/chat` plus `import {should} 'chat'; should();` should be enough to correct the type of `[any object].prorotype.should`, just like this: https://github.com/SukkaW/rollup-plugin-swc/blob/master/test/index.ts

at my device using prototype.should getting error (throw undefined and make all tests failed), instead giving result test failure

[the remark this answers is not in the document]
dimaslanjaka: i dont know, why. But using normal chai API can give result failure correctly instead throw all tests runner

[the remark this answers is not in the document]
dimaslanjaka: emm, i think this error caused by running on monorepo/workspace project. cuz all dependents grouped at single `node_modules`. but, this method easily to sync between packages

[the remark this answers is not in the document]
dimaslanjaka: error says:

cannot read should property 'should' of undefined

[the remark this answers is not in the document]
dimaslanjaka: sometimes got error like:

cannot read property 'eql' ... so on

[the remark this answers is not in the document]
dimaslanjaka: so to prevent useless error, using original chai expect, should API for safety :)

lib/CacheMapper.ts
dimaslanjaka: when removed all size manipulator, will return undefined
image

dimaslanjaka: or using keys counter, this will works like charm
image

lib/CacheMapper.ts
dimaslanjaka: CacheMapper and Cache generic type fixed and supported to integrated with `hexo.locals` check Test CI https://github.com/dimaslanjaka/hexo/actions/runs/4990595464/jobs/8936162528

PR #356, opened 2023-08-28
test: replace nyc with c8

uiolee:
https://github.com/hexojs/hexo/pull/5040


uiolee:
coverage in ci is lower than that in my pc

.nycrc.json
yoshinorin, line 1: IMHO:

We should change the filename from `.nycrc.json` to `.c8rc.json`.

> refs: https://github.com/bcoe/c8/tree/a13584d5be5259ebb6a00455d352c3e8b16006de#cli-options--configuration

uiolee: yes, i think so

PR #384, opened 2023-10-21
fix(permalink): handle invalid permalink setting

stevenjoezhang:
Fix https://github.com/hexojs/hexo/issues/5263

### DIFF
--- a/lib/permalink.ts
+++ b/lib/permalink.ts
@@ -51,7 +51,8 @@ class Permalink {
 
   stringify(data) {
     return this.rule.replace(rParam, (match, name) => {
-      if (['permalink', 'path'].includes(name)) {
+      const descriptor = Object.getOwnPropertyDescriptor(data, name);
+      if (descriptor && typeof descriptor.get === 'function') {
         throw new Error('Invalid permalink setting!');
       }
       return data[name];
--- a/lib/permalink.ts
+++ b/lib/permalink.ts
@@ -50,7 +50,12 @@ class Permalink {
   }
 
   stringify(data) {
-    return this.rule.replace(rParam, (match, name) => data[name]);
+    return this.rule.replace(rParam, (match, name) => {
+      if (['permalink', 'path'].includes(name)) {
+        throw new Error('Invalid permalink setting!');
+      }
+      return data[name];
+    });
   }
 }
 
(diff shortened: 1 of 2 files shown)


D-Sketon (dismissed):

uiolee (dismissed):

lib/permalink.ts
```diff
@@ -50,7 +50,12 @@ class Permalink {
   }
 
   stringify(data) {
-    return this.rule.replace(rParam, (match, name) => data[name]);
+    return this.rule.replace(rParam, (match, name) => {
+      if (['permalink', 'path'].includes(name)) {
+        throw new Error('Invalid permalink setting!');
```
uiolee, line 55: I think this is not the best change, it makes `hexo-util` and `hexo` become coupled. (Although probably no one uses `hexo-util` except `hexo`)

https://github.com/hexojs/hexo-util/blob/ec7fd7cddcaecf8c1c33d34eddd1776bc5e05064/README.md?plain=1#L404-L405

stevenjoezhang: I've updated the patch to disallow the case where the parameter is a getter. Please take a look to see if this is better.

D-Sketon: > I've updated the patch to disallow the case where the parameter is a getter. Please take a look to see if this is better.

This may affect `:tags` and `:categories`, but maybe no one will use these two parameters because they can't be overridden by front-matter

uiolee: Thank you for your work. it looks better than before

uiolee: `:tags` or `:categories` looks like an array

uiolee (approved):

lib/permalink.ts
```diff
@@ -50,7 +50,13 @@ class Permalink {
   }
 
   stringify(data) {
-    return this.rule.replace(rParam, (match, name) => data[name]);
+    return this.rule.replace(rParam, (match, name) => {
+      const descriptor = Object.getOwnPropertyDescriptor(data, name);
+      if (descriptor && typeof descriptor.get === 'function') {
+        throw new Error('Invalid permalink setting!');
```
uiolee, line 56: I think it can throw out the cause of the error, which makes it easier to locate the problem.

e.g.
```js
throw new Error(`xxxxx :${name}`);
```

PR #394, opened 2024-02-02
feat(scripts): use hljs internal method to get aliases

uiolee:
As the title.

This will make it a little bit faster 

### time

time

### checksum

checksum


stevenjoezhang (approved):

stevenjoezhang:
We just need to be mindful of whether this internal method will be modified in the future.

PR #398, opened 2024-02-20
feat(highlight): add an option to switch stripIndent

uiolee:
backwards compatibility

related hexojs/hexo#5422


lib/highlight.ts
```diff
@@ -16,11 +16,14 @@ interface Options {
   mark?: number[];
   tab?: string;
   wrap?: boolean;
+  disableStripIndent?: boolean
```
yoshinorin, line 19: IMHO, I prefer to use `stripIndent` instead of `disableStripIndent` for this naming. Because, other existing boolean options are positive form, but this one is negative form.

yoshinorin (commented):
And... would you please update a README?

https://github.com/hexojs/hexo-util/tree/0cbe3bb6c8bda06a64813630a0dedccdbfc2293e?tab=readme-ov-file#highlightstr-options

https://github.com/hexojs/hexo-util?tab=readme-ov-file#prismhighlightstr-options

TripleCamera:
Hi. How is everything going? It's the end of March already. :eyes:

uiolee:
review pls

yoshinorin (approved):
LGTM 👍

PR #399, opened 2024-02-22
refactor: migrate to ts and remove chai-as-promised

D-Sketon:


lib/full_url_for.ts
```diff
@@ -25,7 +25,7 @@ function fullUrlForHelper(path = '/') {
     path = prettyUrls(path, prettyUrlsOptions);
 
     return path;
-  });
+  }) as string;
```
SukkaW, line 28: The type cast here is unnecessary. Just improve `cache.apply`'s type declaration instead.

SukkaW (changes requested):

SukkaW (approved):

PR #354, opened 2023-08-25
test: improve coverage

D-Sketon:
https://github.com/hexojs/hexo/issues/3984

PR #404, opened 2024-04-21
test(coverage): ignore build_highlight_alias

D-Sketon:
## check list

- [ ] Add test cases for the changes.
- [x] Passed the CI test.

## Additional information
ref: https://github.com/hexojs/hexo-util/pull/354#issuecomment-1695807322

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -1,3 +1,5 @@
 {
-  "all": "true"
+  "include": [
+    "lib/**/*.ts"
+  ]
 }


.c8rc.json
```diff
@@ -1,3 +1,5 @@
 {
-  "all": "true"
```
uiolee, line 2: "all" is required.
Otherwise those files which are not introduced will be ignored. For example, if you delete `test/highlight.spec.ts`, you will find that the coverage reaches 100

PR #414, opened 2025-03-04
refactor: remove alias json for support add custom language support

yutingzhao1991:
## check list

- [x] Passed the CI test.

## Description

use `hljs.getLanguage` for check language support, so I can add new language support with hexo plugin.

## Additional information

for https://github.com/hexojs/hexo/issues/5639


stevenjoezhang:
The `highlight_alias.json` file was introduced almost a decade ago, apparently to address efficiency issues with loading aliases in highlight.js: https://github.com/hexojs/hexo-util/commit/7f6469e1bd83abc56f23414ce92fe84f13bc9a88
 If this issue no longer exists, using `hljs.getLanguage` would indeed be a better option.

SukkaW (changes requested):
What about still preferring the alias, but even if the existing alias doesn't match, then use the `hljs.getLanguage` as a fallback? This prevents unexpected behavior changes.

test/highlight.spec.ts
```diff
@@ -1,8 +1,11 @@
 import chai from 'chai';
 import hljs from 'highlight.js';
+import { solidity } from 'highlightjs-solidity';
 import { encode } from 'html-entities';
 import highlight from '../lib/highlight';
 
+hljs.registerLanguage('solidity', solidity);
```
SukkaW, line 7: This introduces side effects in the unit tests.

SukkaW (changes requested):

yutingzhao1991:
> What about still preferring the alias, but even if the existing alias doesn't match, then use the `hljs.getLanguage` as a fallback? This prevents unexpected behavior changes.

You are right, I already fix it.

btw: many useless commit, use squash merge.

SukkaW (approved):
LGTM now. What do you think? @hexojs/core

yoshinorin (changes requested):
@yutingzhao1991 
Would you please fix lint errors? Other than that, LGTM.

PR #427, opened 2025-07-26
fix: ensure PrismJS works in ESM builds and aliases are resolved

dimaslanjaka:
## check list

- [ ] Add test cases for the changes.
- [x] Passed the CI test.

## Description

This PR resolves PrismJS compatibility issues in ESM environments and improves language alias handling. Key changes:

### ✅ Fixes
- Replaces use of `prismjs/components` (CJS-only) with direct ESM-compatible imports (`prismjs/components/index.js`).
- Uses `PrismCore.languages` instead of relying on CJS metadata for language definitions and aliases.
- Adds fallback logic to dynamically load missing languages using `prismLoadLanguages(...)`.

### 🧠 Enhancements
- Adds a `manualAlias` mapping for common aliases like `js` → `javascript`, `sh` → `bash`, etc.
- Consolidates alias resolution from both `manualAlias` and `Prism.languages.alias`.
- Ensures Prism highlighting still works gracefully even when a language is not preloaded.

## Additional information
- Compatible with both ESM and CJS build setups.
- More robust handling of language codes and aliases, reducing false negatives and errors.
- Better runtime resilience in rendering code blocks.

### DIFF
--- a/lib/prism.ts
+++ b/lib/prism.ts
@@ -1,23 +1,24 @@
+import PrismCore from 'prismjs';
+import prismLoadLanguages from 'prismjs/components/index.js';
 import stripIndent from 'strip-indent';
-import prismLoadLanguages from 'prismjs/components/';
 
 let Prism: typeof import('prismjs') | undefined;
 
-// https://github.com/PrismJS/prism/issues/2145
-import prismComponents from 'prismjs/components';
-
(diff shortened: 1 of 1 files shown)


dimaslanjaka:
@SukkaW @stevenjoezhang need review

PR #431, opened 2025-07-31
feat(util): add cross-platform __dirname and __filename support for ESM

dimaslanjaka:
## Checklist

- [x] Pass the CI tests.

## Description

This PR adds cross-runtime path utilities to improve **compatibility with both CommonJS (CJS) and ECMAScript Modules (ESM)** now and in future Hexo releases.

### Key changes
- **New module**: `lib/cross_dirname.ts` providing `getDirname` and `getFilename` utilities.
- Works seamlessly in **both CJS and ESM environments** without relying on Node.js-specific globals like `__dirname` or `__filename`.
- Uses error stack parsing for platform-independent path resolution.
- Updated **README** with documentation and usage examples.
- Added **unit tests** to ensure consistent behavior in both module systems.

### Why
- Prepares Hexo for **full ESM compatibility** while maintaining support for existing CJS usage.
- Avoids deprecated or environment-specific globals to ensure **future-proof path resolution**.

## Additional information
_No breaking changes. CJS projects can adopt immediately; ESM compatibility is transparent._

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Utilities for [Hexo].
 
 - [Installation](#installation)
 - [Usage](#usage)
+- [Cross __dirname](#crossdirname)
 - [Cache](#cache)
 - [CacheStream](#cachestream)
 - [camelCaseKeys](#camelcasekeysobj-options)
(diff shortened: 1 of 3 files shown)


SukkaW (changes requested):
Cross `__dirname`/`import.meta.dirname` support should probably be implemented by Node.js natively instead. I have had enough of self-baked workaround handling loads of edge cases.

And if this is what it takes to go ESM, let's probably just never go ESM at all.

PR #449, opened 2026-08-12
chore: require Node.js 22.1 or newer

stevenjoezhang:
## What does it do?

- Raises the minimum supported Node.js version to 22.1.0.
- Updates the test matrix to Node.js 22, 24, and 26 across Linux, Windows, and macOS.
- Updates the Node.js type definitions to the Node 22 line.

## Why is it needed?

Node.js 20 has reached end of life. Node.js 22.1.0 is also the first Node 22 release that provides the non-throwing WHATWG `URL.parse()` API, which will allow a focused follow-up to remove the remaining legacy `url.parse()` calls discussed in hexojs/hexo#3846.

Users running Node.js 18 or 20 will need to upgrade before adopting the next release containing this change.

## Checks

- `npm test` on Node.js 26: 257 passing
- `npm run eslint`
- TypeScript build and 257 tests on Node.js 22.1.0


SukkaW (approved):

SukkaW:
Do remember this is a breaking change though.

PR #450, opened 2026-08-13
refactor: migrate to URL.parse()

### LINKED ISSUE #3846: WHATWG URL API performance
Since https://github.com/hexojs/hexo-util/pull/114, we start to mitigate deprecated Legacy URL API to newer WHATWG URL API. Current status of mitigation can be found at [this project board](https://github.com/orgs/hexojs/projects/6).

In order to fix https://github.com/hexojs/hexo/issues/3796#issuecomment-546692793, I mitigate `external_link` filter from Legacy URL API to WHATWG URL API after #3812 & https://github.com/hexojs/hexo-util/pull/119.

But immediately a critical performance issue came up after #3812 is merged (see #3833). @curbengh trying to revert #3812 in #3834) and it seems clear that WHATWG API is the cause.

I have carried out series of test trying to optimize `external_link` filter (#3839). During the research of WHATWG URL API, I have noticed a comment from maintainer of BootstrapCDN: https://github.com/MaxCDN/bootstrapcdn/pull/1391

> While deprecated, it's much cleaner to use this when we have relative URLs. It also appears to be a lot faster.

So I bring up a [benchmark](https://runkit.com/sukkaw/5dc58022d0ba090013cdeae8):

```js
const Benchmark = require('benchmark');
const Suite = new Benchmark.Suite;
const { parse, URL } = require('url');
const urlObj = (str) => {
  try {
    return new URL(str);
  } catch (e) {
    return str;
  }
};
 
// add tests
Suite.add('Legacy URL parse()', () => {
  parse('https://skk.moe/path/to/something');
  parse('/path/to/something');
}).add('WHATWG URL API', () => {
  new URL('https://skk.moe/path/to/something');
  new URL('/path/to/something', 'https://skk.moe/');
}).add('safe WHATWG URL', () => {
  urlObj('https://skk.moe/path/to/something');
  urlObj('/path/to/something');
}).on('cycle', function(event) {
  console.info(String(event.target));
}).on('complete', function() {
  console.info('Fastest is ' + this.filter('fastest').map('name'));
}).run();
```

Result is clear, Legacy URL API is at least 4x faster than WHATWG URL API. And the approach (`urlObj()`) we came up to handle relative URL is even slower:

image

Based on the benchmark, I bring up https://github.com/hexojs/hexo/pull/3839/commits/20d4ae4ab270a6010a8a130fa988f27eccc2d850 in #3839 , avoid `urlObj()` by passing a base to `new URL()` to handle relative path. The generation performance backs to normal after this commit.

IMHO, although WHATWG URL API is powerful and safer, since the Legacy URL API is much more faster and quite useful when handling relative path, we should use WHATWG URL API only when necessary. And `try ... catch (err) ...` is expensive so we should avoid `urlObj()` approach. Takes https://github.com/hexojs/hexo-util/pull/114 as an example, for those function that requires `bind(hexo)`(like `full_url_for()` and `url_for()`), we should pass `hexo.config.url` as a `base `to `new URL()` instead of using `urlObj()`.

stevenjoezhang:
## check list

- [x] Passed the CI test.

## Description

Migrate the remaining legacy `url.parse()` call sites to the non-throwing WHATWG `URL.parse()` API, now that hexo-util requires Node.js 22.1 or newer after #449.

- Parse complete absolute URL inputs in `encodeURL()` and `decodeURL()`, while preserving the fast path for ordinary relative paths.
- Replace `new URL()` exception handling with explicit `null` handling.
- Use WHATWG hostname parsing in `isExternalLink()`, `full_url_for()`, and `url_for()`.
- Treat malformed URLs as fallback input instead of throwing, and support IPv6 site hosts correctly.

Issue resolved: hexojs/hexo#3846.

## Additional information

Local checks:

- TypeScript build and 266 tests on Node.js 22.1.0
- TypeScript build and 266 tests on Node.js 24.15.0
- TypeScript build and 266 tests on Node.js 26.7.0
- `npm run eslint`
- 100% line coverage for all five modified runtime files


SukkaW (approved):
BTW how benchmark shows?

stevenjoezhang:
@SukkaW I benchmarked `hexo-util` before and after this PR, with Node.js v26.7.0 and Apple M1 Pro, 32 GiB RAM. The results:

| Workload | Before | After | Change |
|---|---:|---:|---:|
| `encodeURL()` — absolute URL | 224,474 ops/s | 235,755 ops/s | **+5.0%** |
| `encodeURL()` — relative-path fast path | 1,796,595 ops/s | 1,803,873 ops/s | +0.4% |
| `decodeURL()` — absolute URL | 407,214 ops/s | 444,515 ops/s | **+9.2%** |
| `decodeURL()` — relative-path fast path | 2,569,560 ops/s | 2,558,090 ops/s | −0.4% |
| `isExternalLink()` — cache miss | 541,076 ops/s | 674,181 ops/s | **+24.6%** |
| `full_url_for()` — cache miss | 170,467 ops/s | 190,514 ops/s | **+11.8%** |
| `url_for()` — cache miss | 312,992 ops/s | 365,563 ops/s | **+16.8%** |

The ±0.4% differences in the relative-path fast paths are within normal run-to-run variance, so those paths can be considered unchanged.

D-Sketon (approved):

